PR #2, opened 2020-01-10
Add instance:logs command

Njuelle:
## What does this PR do?
Add `instance:logs` command. 

if no instance is specified with `--instance`, displays the list of all running kuzzle instances 
image

* `$ ./bin/run instance:logs`
* `$ ./bin/run instance:logs --instance <your_running_instance_name>`

(also use `-f` flag to follow logs output)


src/commands/instance/logs.ts
```diff
@@ -0,0 +1,79 @@
...
+        name: 'instance',
+        message: 'On which kuzzle instance do you want to see the logs',
+        type: 'list',
+        choices: instancesList,
+      }])
+      instance = responses.instance!
+    }
+    try {
+      await this.showInstanceLogs(instance, followOption)
+    } catch {
```
Yoann-Abbes: Should be on the line below according to our standards
Same remark for others try/catch

src/commands/instance/logs.ts
```diff
@@ -0,0 +1,79 @@
...
+      containersListProcess = await execa('docker', ['ps', '--format', '"{{.Names}}"'])
+    } catch {
+      this.warn('Something went wrong while getting kuzzle running instances list')
+      return
+    }
+
+    const containersList: string[] = containersListProcess.stdout.replace(/"/g, '').split('\n')
+
+    return containersList.filter(containerName => {
+      if (containerName.includes('kuzzle')
```
Yoann-Abbes:
```suggestion
      if ( containerName.includes('kuzzle')
```

PR #6, opened 2020-02-10
Add es:get and es:insert commands

Aschen:
## What does this PR do?

Adds commands to directly get or insert document to Elasticsearch.  
Also adds a command to list ES indexes

#### `kourou es:get INDEX ID`

Get a document from ES

```
USAGE
  $ kourou es:get INDEX ID

ARGUMENTS
  INDEX  ES Index name
  ID     Document ID

OPTIONS
  -h, --host=host  [default: localhost] Elasticsearch server host
  -p, --port=port  [default: 9200] Elasticsearch server port
  --help           show CLI help
```

#### `kourou es:insert INDEX`

Insert a document directly into ES (will replace if exists)

```
USAGE
  $ kourou es:insert INDEX

ARGUMENTS
  INDEX  ES Index name

OPTIONS
  -h, --host=host  [default: localhost] Elasticsearch server host
  -p, --port=port  [default: 9200] Elasticsearch server port
  --body=body      [default: {}] Document body in JSON
  --help           show CLI help
  --id=id          Document ID
```

#### `kourou es:list-index`

List available ES indexes

```
USAGE
  $ kourou es:list-index

OPTIONS
  -g, --grep=grep  Match output with pattern
  -h, --host=host  [default: localhost] Elasticsearch server host
  -p, --port=port  [default: 9200] Elasticsearch server port
  --help           show CLI help
```

### Other changes

  - Ocliff silently catch errors so I had wrapper to ensure we display them to the user


src/commands/es/get.ts
```diff
@@ -0,0 +1,58 @@
+import { flags } from '@oclif/command'
+import { Kommand } from '../../common'
+import { Client } from '@elastic/elasticsearch'
+
+export default class EsGet extends Kommand {
+  static description = 'Get a document from ES'
+
+  static flags = {
```
rolljee: Since this is a TS file, this should expose an interface for flag named `IEsGetFlag` ?

src/commands/es/insert.ts
```diff
@@ -0,0 +1,65 @@
+import { flags } from '@oclif/command'
+import { Kommand } from '../../common'
+import { Client } from '@elastic/elasticsearch'
+
+export default class EsInsert extends Kommand {
+  static description = 'Insert a document directly into ES (will replace if exists)'
+
+  static flags = {
```
rolljee: Since this is a TS file, this should expose an interface for flag named `IEsInsertFlag` ?

src/commands/es/list-index.ts
```diff
@@ -0,0 +1,60 @@
+import { flags } from '@oclif/command'
+import { Kommand } from '../../common'
+import { Client } from '@elastic/elasticsearch'
+
+export default class EsListIndex extends Kommand {
+  static description = 'List available ES indexes'
+
+  static flags = {
```
rolljee: Since this is a TS file, this should expose an interface for flag named `IEsInsertIndexFlag` ?

src/commands/es/get.ts
Aschen: I don't understand because I don't have much experience in TS, I use this `static` definition because it's the way oclif will infer command flags https://oclif.io/docs/flags

src/commands/es/insert.ts
Aschen: Ditto

src/commands/es/list-index.ts
Aschen: Ditto

src/commands/es/get.ts
rolljee:
```javascript
interface LabeledValue {
  label: string;
}

function printLabel(labeledObj: LabeledValue) {
  console.log(labeledObj.label);
}

let myObj = { size: 10, label: "Size 10 Object" };
printLabel(myObj);
```

Interface just serve the purpose of defining the object structure you are manipulating, it does not take any default value nor any description.

```javascript
static flags = {
    help: flags.help(),
    host: flags.string({
        char: 'h',
        description: 'Elasticsearch server host',
        default: process.env.KUZZLE_HOST || 'localhost',
    }),
    port: flags.string({
        char: 'p',
        description: 'Elasticsearch server port',
        default: process.env.KUZZLE_PORT || '9200',
    }),
}
```

will be written in TS (using ocliff type definition)

```ts
import { flags } from '@oclif/command'
import { IOptionFlag, IBooleanFlag } from '@oclif/parser/lib/flags'

export interface IFlags {
    help: IBooleanFlag<void>;
    host: IOptionFlag<string>;
    port: IOptionFlag<string>;
}


export class myclass {
    static description = 'Get a document from ES'
    static flags: IFlags = {
        help: flags.help(),
        host: flags.string({
            char: 'h',
            description: 'Elasticsearch server host',
            default: process.env.KUZZLE_HOST || 'localhost',
        }),
        port: flags.string({
            char: 'p',
            description: 'Elasticsearch server port',
            default: process.env.KUZZLE_PORT || '9200',
        }),
    }
}
```

src/commands/es/insert.ts
rolljee: Ditto

src/commands/es/list-index.ts
rolljee: Ditto

src/commands/es/get.ts
Aschen: Ok I see but it seems redundant to me to redeclare the types because they already are automatically inferred when the static `flags` variable is declared.  
```
static flags: IFlags = {
        help: flags.help(),
        host: flags.string(...),
  }

flags.help.length // type error from TS because flags.help is a boolean
```

rolljee: ocliff and typescript doesn't serve the same purpose that's why it seem's redundant

- ocliff provide a command line tool helper
- typescript provide the typing that will ensure you are using the proper type and properties for the interface you are using.

Aschen: Yes but my point is oclif methods return strong typescript types

Aschen:
We talked with @rolljee and I agree that strong typing could be useful to accelerate development and to avoid mistakes but like everything it comes with a cost and this cost could be expensive because you will have to define interfaces for every object you use and also access the types defined by the library.

So for the CLI, I don't want to have a painful development journey since it's a fairly simple project and each command is more or less independent so the complexity is very low.

rolljee:
> We talked with @rolljee and I agree that strong typing could be useful to accelerate development and to avoid mistakes but like everything it comes with a cost and this cost could be expensive because you will have to define interfaces for every object you use and also access the types defined by the library.
> 
> So for the CLI, I don't want to have a painful development journey since it's a fairly simple project and each command is more or less independent so the complexity is very low.

For the sake of this PR adding interface is not mandatory, everithing else is fine :+1:

README.md
```diff
@@ -74,7 +77,7 @@ By environment variables:
 
 ## `kourou api-key:create USER`
 
-Creates a new API Key for a user.
+Creates a new API Key for an user
```
scottinet: You need to import fixes from #7 (applies to all README fixes)

README.md
```diff
@@ -215,7 +218,65 @@ OPTIONS
   --username=username  [default: anonymous] Kuzzle user
 ```
 
-_See code: [src/commands/document/get.ts](https://github.com/kuzzleio/kourou/blob/v0.6.0/src/commands/document/get.ts)_
+_See code: [src/commands/document/get.ts](https://github.com/kuzzleio/kourou/blob/v0.7.0/src/commands/document/get.ts)_
+
+## `kourou es:get INDEX ID`
+
+Get a document from ES
```
scottinet:
```suggestion
Gets a document from ES
```

README.md
```diff
@@ -215,7 +218,65 @@ OPTIONS
...
+  -h, --host=host  [default: localhost] Elasticsearch server host
+  -p, --port=port  [default: 9200] Elasticsearch server port
+  --help           show CLI help
+```
+
+_See code: [src/commands/es/get.ts](https://github.com/kuzzleio/kourou/blob/v0.7.0/src/commands/es/get.ts)_
+
+## `kourou es:insert INDEX`
+
+Insert a document directly into ES (will replace if exists)
```
scottinet:
```suggestion
Inserts a document directly into ES (will replace if exists)
```

scottinet: Also, why "directly"? Does this mean that there won't be real-time notifications triggered?

README.md
```diff
@@ -215,7 +218,65 @@ OPTIONS
...
+  --body=body      [default: {}] Document body in JSON
+  --help           show CLI help
+  --id=id          Document ID
+```
+
+_See code: [src/commands/es/insert.ts](https://github.com/kuzzleio/kourou/blob/v0.7.0/src/commands/es/insert.ts)_
+
+## `kourou es:list-index`
+
+List available ES indexes
```
scottinet:
```suggestion
Lists available ES indexes
```

src/commands/es/get.ts
```diff
@@ -0,0 +1,58 @@
+import { flags } from '@oclif/command'
+import { Kommand } from '../../common'
+import { Client } from '@elastic/elasticsearch'
+
+export default class EsGet extends Kommand {
+  static description = 'Get a document from ES'
```
scottinet:
```suggestion
  static description = 'Gets a document from ES'
```

src/commands/es/insert.ts
```diff
@@ -0,0 +1,65 @@
+import { flags } from '@oclif/command'
+import { Kommand } from '../../common'
+import { Client } from '@elastic/elasticsearch'
+
+export default class EsInsert extends Kommand {
+  static description = 'Insert a document directly into ES (will replace if exists)'
```
scottinet:
```suggestion
  static description = 'Inserts a document directly into ES (will replace if exists)'
```

src/commands/es/list-index.ts
```diff
@@ -0,0 +1,60 @@
+import { flags } from '@oclif/command'
+import { Kommand } from '../../common'
+import { Client } from '@elastic/elasticsearch'
+
+export default class EsListIndex extends Kommand {
+  static description = 'List available ES indexes'
```
scottinet:
```suggestion
  static description = 'Lists available ES indexes'
```

src/commands/collection/dump.ts
```diff
@@ -26,6 +26,15 @@ export default class CollectionDump extends Kommand {
   ]
 
   async run() {
+    try {
```
scottinet: (nitpicking / open to debate) there are so many of these everywhere, and it's so generic, that I think that this `run` command should be put in the `Kommand` parent class instead. 

If you do so, I would reverse the names because the safe runner is this one, not the one invoked by it:
* the Kommand class should have a `runSafe`method with almost the same code as the one I'm highlighting in this comment (only change is it would invoke "run" instead of "runSafe" obviously)
* the Kommand class should have  a `run` method throwing a "Not implemented" error when invoked
* other classes should overload the "run" command as they see fit
* to use these functions, we can chose to explicitly call "run" or "runSafe" (I understand that only the latter shall be invoked but, you know...)

README.md
Aschen: Yes, because method under the `es` namespace send directly a request to Elasticsearch, it won't pass through Kuzzle

src/commands/collection/dump.ts
Aschen: The `run` command is called internally by oclif so I will put the `try.. catch` code inside the `Kommand` class and call the class `runUnsafe` method so it will avoid code duplication.

But I will apply these change in a separate PR So I can correct all the new method at once :sweat_smile:

PR #18, opened 2020-03-16
Adds more login time to allow dumping large collections

Aschen:
## What does this PR do?

If the dumping action takes too long, the SDK will not be connected anymore.  
This PR change the token validity to 10 minutes instead of 60 sec.  
The token will also be refreshed after 8 minutes

### DIFF
--- a/src/support/kuzzle.ts
+++ b/src/support/kuzzle.ts
@@ -65,7 +65,7 @@ export class KuzzleSDK {
         password: this.password,
       }
 
-      await this.sdk.auth.login('local', credentials, '60s')
+      await this.sdk.auth.login('local', credentials, '2h')
     }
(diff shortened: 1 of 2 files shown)


src/support/kuzzle.ts
```diff
@@ -4,6 +4,8 @@ import chalk from 'chalk'
 // tslint:disable-next-line
 const { Http, Kuzzle } = require('kuzzle-sdk')
 
+const EIGHT_MINUTES = 5 * 60 * 1000;
```
benoitvidis: Is it really eight minutes?

Aschen: Ohh.. Good catch :sweat_smile:

PR #21, opened 2020-03-24
Adds document:search

Aschen:
## What does this PR do?

Adds a `document:search` command.  
This command take the `--query` and `--sort`  flags.  
It's also possible to edit the request before being sent to Kuzzle and manually with the `--editor` option.

[asciicast](https://asciinema.org/a/ZTAdVlEFnmzxFkKmRBWp1HJBS)

```
USAGE
  $ kourou document:search INDEX COLLECTION

ARGUMENTS
  INDEX       Index name
  COLLECTION  Collection name

OPTIONS
  -h, --host=host      [default: localhost] Kuzzle server host
  -p, --port=port      [default: 7512] Kuzzle server port
  --editor             Open an editor (EDITOR env variable) to edit the request before sending
  --from=from          Optional offset
  --help               show CLI help
  --password=password  Kuzzle user password
  --query=query        [default: {}] Query in JS or JSON format.
  --scroll=scroll      Optional scroll TTL
  --size=size          Optional page size
  --sort=sort          [default: {}] Sort in JS or JSON format.
  --ssl                Use SSL to connect to Kuzzle
  --username=username  [default: anonymous] Kuzzle username (local strategy)

EXAMPLES
  kourou document:search iot sensors --query '{ term: { name: "corona" } }'
  kourou document:search iot sensors --editor
```
### Other changes

 - also allow kourou query to be edited with `--editor` flag

### DIFF
--- a/src/support/editor.ts
+++ b/src/support/editor.ts
@@ -40,7 +40,7 @@ export class Editor {
       this.tmpFile = tmp.fileSync({ postfix: '.json' })
     }
 
-    fs.writeFileSync(this.tmpFile.name, this._content);
+    fs.writeFileSync(this.tmpFile.name, this._content)
   }
 
   run() {
--- a/src/support/editor.ts
+++ b/src/support/editor.ts
@@ -7,7 +7,7 @@ export interface IEditorParams {
   json?: boolean;
 }
 
-export default class Editor {
+export class Editor {
   private _content: string;
   private tmpFile: string;
   private defaultEditor: string;
@@ -45,7 +45,8 @@ export default class Editor {
 
     const response = spawnSync(
       editor,
-      [this.tmpFile]);
+      [this.tmpFile],
+      { stdio: "inherit" });
 
     if (response.status !== 0) {
       console.error(response.stdout)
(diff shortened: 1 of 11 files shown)


src/support/editor.ts
```diff
@@ -0,0 +1,84 @@
...
+      editor,
+      [this.tmpFile],
+      { stdio: 'inherit' })
+
+    if (response.status !== 0) {
+      // eslint-disable-next-line
+      console.error(response.stdout)
+      // eslint-disable-next-line
+      console.error(response.stderr)
+      throw new Error(`Unable to open editor "${editor}": ${response.error}.\nPlease set EDITOR environment variable.`)
```
scottinet: Temporary file is not removed if error is thrown

Aschen: I use the `tmp` module now. Thanks!

src/support/editor.ts
```diff
@@ -0,0 +1,71 @@
...
+        this._content = '{\n\n}'
+      }
+      else {
+        this._content = JSON.stringify(JSON.parse(this._content), null, 2)
+      }
+
+      this.tmpFile = tmp.fileSync({ postfix: '.json' })
+    }
+    else {
+      this.tmpFile = tmp.fileSync({ postfix: '.json' })
```
scottinet: Superfluous else branch: the tmpFile variable content doesn't depend on the result of the if condition

Aschen: Bad copy/paste ..

PR #24, opened 2020-03-25
Fix login timeout on dump/restore

Aschen:
## What does this PR do?

Dump and restore index can be very long so the token will be periodically refreshed.

I had to all commands to add a `cleanInterval` after the command has run

### Other changes

 - catch error and continue to process other collection when an export/import fail

### DIFF
--- a/src/support/kuzzle.ts
+++ b/src/support/kuzzle.ts
@@ -4,7 +4,7 @@ import chalk from 'chalk'
 // tslint:disable-next-line
 const { Http, WebSocket, Kuzzle } = require('kuzzle-sdk')
 
-const ONE_SECOND = 60 * 1000;
+const ONE_SECOND = 60 * 1000
 
 export const kuzzleFlags = {
   host: flags.string({
@@ -60,11 +60,11 @@ export class KuzzleSDK {
   }
 
   public async init(log: any) {
-    const protocolClass = this.protocol === 'ws'
+    const ProtocolClass = this.protocol === 'ws'
       ? WebSocket
       : Http
 
-    this.sdk = new Kuzzle(new protocolClass(this.host, {
+    this.sdk = new Kuzzle(new ProtocolClass(this.host, {
       port: this.port,
       sslConnection: this.ssl,
     }))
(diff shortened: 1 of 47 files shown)


src/support/kuzzle.ts
```diff
@@ -41,24 +43,33 @@ export class KuzzleSDK {
...
+  private refreshTimer?: NodeJS.Timeout;
 
   constructor(options: any) {
     this.host = options.host
     this.port = parseInt(options.port, 10)
     this.ssl = options.ssl || this.port === 443
     this.username = options.username
     this.password = options.password
-    this.loginTTL = options.loginTTL || '10s'
+    this.refreshLogin = options.refreshLogin || false
```
scottinet: Why should this be an option?

As a user, I expect the CLI to take care of how the authentication token is created and refreshed.

If any command runs longer than the default login TTL (which is 90s), then I think that Kourou should automatically renew its token until the command ends.

Also, renewing every second or so seems a bit too often to me. If you have a login TTL of 90s, why not refreshing every 80s or so (10s to account for network latency seems largely enough)

What do you think?

src/support/kuzzle.ts
```diff
@@ -68,8 +79,28 @@ export class KuzzleSDK {
...
+      }
+
+      log(chalk.green(`[ℹ] Loggued as ${this.username} (refreshLogin: ${this.refreshLogin}).`))
+    }
+  }
+
+  disconnect() {
+    this.sdk.disconnect()
+
+    if (this.refreshTimer) {
```
scottinet: (nitpicking) the behavior of `clearInterval` on null or undefined vaues is a no-op, so this if-condition is unnecessary

src/support/kuzzle.ts
Aschen: Yes you're right, I'm gonna refresh it all the time if needed :+1:

src/support/kuzzle.ts
Aschen: I can't remove the condition otherwise Typescript will complain :expressionless:

PR #29, opened 2020-04-06
Fix collection list pagination

Aschen:
## What does this PR do?

Retrieve every collections when dumping an index


Aschen:
I will update the SDK instead

PR #30, opened 2020-04-09
Fix token refresh

Aschen:
## What does this PR do?

Fix token refresh interval :expressionless: 

### Other changes

 - always dump index in a directory having the same name

### DIFF
--- a/src/support/kuzzle.ts
+++ b/src/support/kuzzle.ts
@@ -4,7 +4,7 @@ import chalk from 'chalk'
 // tslint:disable-next-line
 const { Http, WebSocket, Kuzzle } = require('kuzzle-sdk')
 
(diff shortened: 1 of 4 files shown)


src/support/kuzzle.ts
```diff
@@ -4,7 +4,7 @@ import chalk from 'chalk'
 // tslint:disable-next-line
 const { Http, WebSocket, Kuzzle } = require('kuzzle-sdk')
 
-const SECOND = 60 * 1000
+const SECOND = 1000
```
scottinet, line 7: I guess I did not pay attention to this either when I reviewed this feature :expressionless:

Aschen: :sweat_smile:

PR #33, opened 2020-04-13
Add vault:decrypt

Aschen:
## What does this PR do?

This PR adds the command `vault:decrypt` who decrypt an entire encrypted secrets file.

### How should this be manually tested?

````
$ export KUZZLE_VAULT_KEY=password
$ kourou vault:add secrets.enc.json aws.key "aws secret key"
$ kourou vault:add secrets.enc.json aws.id "aws secret key"
# decrypt the file
$ kourou vault:decrypt secrets.enc.json
```

### Other changes

 - use the latest version of the Vault (2.0.0) written in Typescript

PR #34, opened 2020-04-13
Add file encrypt

Aschen:
## What does this PR do?

:warning: Depends on https://github.com/kuzzleio/kourou/pull/33

Adds commands to encrypt and decrypt arbitrary files:
```
$ kourou file:encrypt
$ kourou file:decrypt
$ kourou file:test
```

Those files are encrypted with the same crypto routines as for the Vault.

```bash
$ export KUZZLE_VAULT_KEY=password
$ kourou-dev file:encrypt package.json
$ kourou-dev file:test package.json.enc
$ kourou-dev file:decrypt package.json -o decrypted.json
```


src/commands/file/encrypt.ts
```diff
@@ -0,0 +1,79 @@
...
+    try {
+      content = fs.readFileSync(args.file, 'utf8')
+    }
+    catch (error) {
+      throw new Error(`Cannot read file "${args.file}": ${error.message}`)
+    }
+
+    const encryptedContent = cryptonomicon.encryptString(content)
+
+    fs.writeFileSync(outputFile, encryptedContent)
```
Yoann-Abbes, line 75: shouldnt be in a try/catch ?

src/commands/file/decrypt.ts
```diff
@@ -0,0 +1,79 @@
...
+
+    let encryptedContent
+    try {
+      encryptedContent = fs.readFileSync(args.file, 'utf8')
+    }
+    catch (error) {
+      throw new Error(`Cannot read encrypted content from file "${args.file}": ${error.message}`)
+    }
+
+    const content = cryptonomicon.decryptString(encryptedContent)
```
Leodau, line 73: Shouldn't this be in a try catch as well?

Aschen: Everything is inside a `try...catch` block since the `runSafe` method is called inside one https://github.com/kuzzleio/kourou/blob/master/src/common.ts#L29

src/commands/file/encrypt.ts
Aschen: Everything is inside a `try...catch` block since the `runSafe` method is called inside one https://github.com/kuzzleio/kourou/blob/master/src/common.ts#L29

src/commands/vault/decrypt.ts
```diff
@@ -39,20 +38,24 @@ export class VaultDecrypt extends Kommand {
...
     if (userFlags['output-file']) {
       outputFile = userFlags['output-file']
     }
 
+    if (fs.existsSync(outputFile)) {
+      if (!userFlags.force) {
+        throw new Error(`Output file "${outputFile}" already exists. Use -f flag to overwrite it.`)
+      }
+    }
+
```
alexandrebouthinon, line 58: NP: Is this cannot be simplified? Like this:
```suggestion
    if (fs.existsSync(outputFile) && !userFlags.force) {
      throw new Error(`Output file "${outputFile}" already exists. Use -f flag to overwrite it.`)
    }

```

src/commands/file/encrypt.ts
```diff
@@ -0,0 +1,79 @@
...
+    let outputFile = `${args.file}.enc`
+    if (userFlags['output-file']) {
+      outputFile = userFlags['output-file']
+    }
+
+    if (fs.existsSync(outputFile)) {
+      if (!userFlags.force) {
+        throw new Error(`Output file "${outputFile}" already exists. Use -f flag to overwrite it.`)
+      }
+    }
```
alexandrebouthinon, line 57: NP: Is this cannot be simplified? Like this:
```suggestion
    if (fs.existsSync(outputFile) && !userFlags.force) {
      throw new Error(`Output file "${outputFile}" already exists. Use -f flag to overwrite it.`)
    }
```

src/commands/file/decrypt.ts
```diff
@@ -0,0 +1,79 @@
...
+    let outputFile = `${args.file.replace('.enc', '')}`
+    if (userFlags['output-file']) {
+      outputFile = userFlags['output-file']
+    }
+
+    if (fs.existsSync(outputFile)) {
+      if (!userFlags.force) {
+        throw new Error(`Output file "${outputFile}" already exists. Use -f flag to overwrite it.`)
+      }
+    }
```
alexandrebouthinon, line 57: NP: Is this cannot be simplified? Like this:
```suggestion
    if (fs.existsSync(outputFile) && !userFlags.force) {
      throw new Error(`Output file "${outputFile}" already exists. Use -f flag to overwrite it.`)
    }
```

PR #38, opened 2020-04-24
Add command to spot differences between kuzzlerc files

Aschen:
## What does this PR do?

Adds a command to spot differences between kuzzlerc files to ensure configuration consistency between environments.

```
kourou-dev config:diff config/local/kuzzlerc config/production/kuzzlerc
 
 🚀 Kourou - Returns differences between two Kuzzle configuration files (kuzzlerc)
 
 [ℹ] Found key differences between the provided configurations. In the second file:
  - key "plugins.omniscient.s3BucketsRegion" was added
  - key "plugins.sentry.dsn" was added
  - key "plugins.sentry.excludeIntegrations" was added
  - key "plugins.sentry.dsn" is "production" and was "local" 
```

kouro-config-diff

### DIFF
--- a/src/commands/config/key-diff.ts
+++ b/src/commands/config/key-diff.ts
@@ -5,11 +5,11 @@ import stripComments from 'strip-json-comments'
 
 import { Kommand } from '../../common'
 
-export class ConfigDiff extends Kommand {
+export class ConfigKeyDiff extends Kommand {
   static description = 'Returns differences between the keys of two Kuzzle configuration files (kuzzlerc)'
 
   static examples = [
-    'kourou config:diff config/local/kuzzlerc config/production/kuzzlerc'
+    'kourou config:key-diff config/local/kuzzlerc config/production/kuzzlerc'
   ]
 
   static flags = {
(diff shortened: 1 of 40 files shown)


src/commands/config/key-diff.ts
```diff
@@ -0,0 +1,87 @@
...
+          ar.concat(ar2)
+          changes[[...path, key].join('.')] = 'removed'
+        }
+      }
+
+      for (const [key, value] of Object.entries(object)) {
+        if (base[key] === undefined) {
+          changes[[...path, key].join('.')] = 'added'
+        }
+        else if (!_.isEqual(value, base[key]) && _.isObject(value) && _.isObject(base[key])) {
```
scottinet, line 77: Keys added/removed are fine but reading the description I was expecting this command to spot differences in values too. And I think this would be very useful for instance when doing support, to check what changes between different environments.

Aschen: Yes good idea, I'm gonna add this feature too!

PR #40, opened 2020-04-30
Add user impersonation

Aschen:
## What does this PR do?

:warning: depends on #39 

Adds the `--as` flag to impersonate users.  

Under the hood it generate an API key for the user, execute the command and then delete the API key.

```bash
$ kourou query auth:getCurrentUser --as gordon --username admin --password admin
 
 🚀 Kourou - Executes an API query.
 
 [ℹ] Connecting to http://localhost:7512 ...
 [ℹ] Impersonate user "gordon"

[...]
```

```bash
$ kourou query security:createUser --body '{"content":{"profileIds":["default"]},"credentials":{ "local": {"username": "gordon", "password": "password"}}}' -a _id=gordon

$ kourou query auth:getCurrentUser --as gordon
```


README.md
```diff
@@ -240,14 +265,15 @@ ARGUMENTS
...
+  --as=as                  Impersonate a user
   --batch-size=batch-size  [default: 2000] Maximum batch size (see limits.documentsFetchCount config)
   --editor                 Open an editor (EDITOR env variable) to edit the query before sending
   --help                   show CLI help
   --host=host              [default: localhost] Kuzzle server host
   --password=password      Kuzzle user password
   --path=path              Dump root directory
   --port=port              [default: 7512] Kuzzle server port
-  --protocol=protocol      [default: websocket] Kuzzle protocol (http or websocket)
+  --protocol=protocol      [default: ws] Kuzzle protocol (http or websocket)
```
scottinet, line 276: The list of available values need to be updated (for all commands) : `http or websocket` => `http or ws`

src/support/kuzzle.ts
```diff
@@ -88,11 +98,51 @@ export class KuzzleSDK {
...
+  public async impersonate(userKuid: string, callback: Function) {
+    const currentToken = this.sdk.jwt
+
+    let apiKey: any
+
+    try {
+      const apiKey = await this.security.createApiKey(
+        userKuid,
+        'Kourou impersonation token',
+        { expiresIn: '1d', refresh: false })
```
scottinet, line 125: 1d seems a bit too large to my taste... What about 1h instead?

Aschen: Actually I don't think it really matter because the API key is deleted.  
I put `1d` in the case of a user running a long command like export or import but I don't think they will impersonate someone.

I will put 2h

PR #41, opened 2020-05-04
Add sdk:execute to execute arbitrary code

Aschen:
## What does this PR do?

:warning: Depends on https://github.com/kuzzleio/kourou/pull/39

Adds a method to execute arbitrary code in the context of the SDK.

```
$ kourou sdk:execute --code 'const now = await sdk.server.now(); return new Date(now);'
 
 🚀 Kourou - Executes arbitrary code with a connected SDK
 
 [ℹ] Connecting to http://localhost:7512 ...
 [✔] Successfully executed SDK code
 "2020-05-04T08:02:48.690Z"
```

User can also provide variable which will be injected in the code:

```
$ kourou sdk:execute --code 'return await sdk.index.exists(index)' --var 'index="nyc-open-data"'
```

If an error occur, the code is printed for debug:
```
$ kourou sdk:execute --code 'return await sdk.index.exists(foobar)' -v 'index="nyc-open-data"'
 
 🚀 Kourou - Executes arbitrary code.
 
 [ℹ] Connecting to http://localhost:7512 ...
 [X] Error when executing SDK code: foobar is not defined
 
(async () => {
  try {
    let index = "nyc-open-data";
    let collection = 42;
    return await sdk.index.exists(foobar)
  }
  catch (error) {
    userError = error
  }
})();
```

```
$ kourou sdk:execute --code 'const now = await sdk.server.now(); return new Date(now);'
```

### Other changes

 - `kourou query` is now `kourou sdk:query`


src/commands/sdk/execute.ts
```diff
@@ -0,0 +1,90 @@
...
+import { kuzzleFlags } from '../../support/kuzzle'
+
+class SdkQuery extends Kommand {
+  public static description = `
+Executes arbitrary code.
+
+Code Execution
+
+  provided code will be executed in an async method
+  you can access an connected and authenticated SDK with the "sdk" variable
```
Aschen, line 14:
```suggestion
  you can access a connected and authenticated SDK with the "sdk" variable
```

src/commands/sdk/execute.ts
```diff
@@ -0,0 +1,106 @@
...
+  async runSafe() {
+    // try to read stdin
+    const stdin = await this.fromStdin()
+
+    if (stdin && this.flags.editor) {
+      throw new Error('Unable to use flag --editor when reading from STDIN')
+    }
+
+    let code: any = stdin || this.flags.code
+    let userError: any
```
Njuelle, line 63: NIT: Couldn't we cast him as `Error` to avoid another "any"?

scottinet:
Can you provide me with a use case for that one?

Aschen:
> Can you provide me with a use case for that one?
> I can't seem to find one where this feature would be useful.

For example it happen me quite often to create multiple documents with slightly changes (test a search query, one liner to generate document for the `How should this be manually tested` section, etc)
If I want to generate 10 documents with a different `age` and `name`, I have two options:
 - repeat 10 times the kourou command `kourou document:create ...` with bash (I never remember the syntax)
 - use `document:mCreate` with kourou but I will have to create the body by myself

With this command, it's very easy:
```
$ kourou-dev sdk:execute --code 'for(let i =10; i--;) { await sdk.document.create("index", "collection", { age: 18+i, name: `user-${i}` }); }'
```

Another example, before I used to have a `test-sdk.js` file in almost every project because sometime I just want to test some snippet with the SDK.
With this command I will be able to do it without the hassle of initializing, connecting, authenticating the SDK.

```
let results = await sdk.document.search(customer, 'collection', {}, { size: 1000 })
const docs = []
for (const doc of results.hits) {
  if (doc._source.expiresAt >= Date.now()) {
    docs.push({ _id: doc._id, body: { timeout: true } })
  }
}
await sdk.document.mUpdate(customer, 'collection', docs)
```

Then I can execute my snippet with my target environment and authentication. I can also pass variables or even open the editor to add a `console.log` if I want to debug
```
$ kourou sdk:execute -v customer=customer-worksite-foobar < snippet.js
```

scottinet:
I didn't see this that way, but this looks very nice indeed :+1:

README.md
```diff
@@ -727,13 +729,112 @@ OPTIONS
...
+  --protocol=protocol  [default: ws] Kuzzle protocol (http or websocket)
+  --ssl                Use SSL to connect to Kuzzle
+  --username=username  [default: anonymous] Kuzzle username (local strategy)
+```
+
+_See code: [src/commands/role/export.ts](https://github.com/kuzzleio/kourou/blob/v0.11.0/src/commands/role/export.ts)_
+
+## `kourou role:import PATH`
+
+Import roles
```
scottinet, line 756:
```suggestion
Imports roles.
```

README.md
```diff
@@ -727,13 +729,112 @@ OPTIONS
...
+DESCRIPTION
+  Executes arbitrary code.
+
+  Code Execution
+
+     provided code will be executed in an async method
+     you can access a connected and authenticated SDK with the "sdk" variable
+     templated variable passed as the command arguments are also accessible within the same name
+     return value will be printed on the standard output (eg: 'return await sdk.server.now();')
+     error will be catched and printed on the error output (eg: 'throw new Error("failure");')
```
scottinet, line 808:
```suggestion
     Provided code will be executed in an async method.
     You can access a connected and authenticated SDK with the "sdk" variable.
     Templated variable passed as the command arguments are also accessible within the same name.
     Returned value will be printed on the standard output (e.g. 'return await sdk.server.now();')
     Errors will be caught and printed on the error output (e.g. 'throw new Error("failure");').
```

src/commands/sdk/execute.ts
```diff
@@ -0,0 +1,114 @@
...
+  public static description = `
+Executes arbitrary code.
+
+Code Execution
+
+  provided code will be executed in an async method
+  you can access a connected and authenticated SDK with the "sdk" variable
+  templated variable passed as the command arguments are also accessible within the same name
+  return value will be printed on the standard output (eg: 'return await sdk.server.now();')
+  error will be catched and printed on the error output (eg: 'throw new Error("failure");')
```
scottinet, line 17:
```suggestion
  Provided code will be executed in an async method.
  You can access a connected and authenticated SDK with the "sdk" variable.
  Templated variable passed as the command arguments are also accessible within the same name.
  Returned value will be printed on the standard output (e.g. 'return await sdk.server.now();').
  Errors will be caught and printed on the error output (e.g. 'throw new Error("failure");').
```

src/commands/sdk/execute.ts
```diff
@@ -0,0 +1,114 @@
...
+    }),
+    ...kuzzleFlags,
+  };
+
+  async runSafe() {
+    // try to read stdin
+    const stdin = await this.fromStdin()
+
+    if (stdin && this.flags.editor) {
+      throw new Error('Unable to use flag --editor when reading from STDIN')
```
scottinet, line 59:
```suggestion
      throw new Error('Cannot use --editor when reading from STDIN')
```

PR #43, opened 2020-05-13
Add realtime subscription support

Aschen:
## What does this PR do?

This PR adds the command `subscribe` to subscribe to realtime notifications.

```
$ kourou-dev subscribe iot-data sensors --filters '{ range: { temperature: { gt: 0 } } }'
 
 🚀 Kourou - Subscribes to realtime notifications
 
 [ℹ] Connecting to ws://localhost:7512 ...
 [ℹ] Waiting for realtime notifications on "iot-data":"sensors" ...
 [ℹ] New notification
 {
  "_id": null,
  "_source": {
    "temperature": 42,
    "_kuzzle_info": {
      "author": "-1",
      "createdAt": 1589346490187
    }
  }
}
```

I didn't add functional tests because I couldn't figure a way to execute them.. (ie: exit the program after receiving the notification)

```
$ kourou subscribe iot-data sensors --filters '{ range: { temperature: { gt: 0 } } }'
# In another terminal
$ kourou sdk:query realtime:publish -i iot-data -c sensors --body '{ temperature: 42 }'
```

### Other changes
 
 - add `beforeConnect` hook and `this.sdkOptions` to the `Kommon` class  
 - add option to keep the connection alive for `sdk:execute` (allows to have realtime notifications)
 - allow to display a specific property from the response of `sdk:query`


src/commands/sdk/query.ts
```diff
@@ -99,10 +111,14 @@ Other
 
     const response = await this.sdk?.query(request)
 
-    this.log(JSON.stringify(response, null, 2))
+    const display = this.flags.display === ''
+      ? response
+      : _.get(response, this.flags.display)
```
rolljee, line 116: it's usualy better pratice to import lodash as so `import get from 'lodash/get`

src/commands/subscribe.ts
```diff
@@ -0,0 +1,99 @@
+import { flags } from '@oclif/command'
+import * as _ from 'lodash'
```
rolljee, line 2: Same lodash comment

src/commands/subscribe.ts
```diff
@@ -0,0 +1,99 @@
...
+      {
+        scope: this.flags.scope,
+        users: this.flags.users,
+        volatile: this.parseJs(this.flags.volatile),
+      })
+
+    this.logInfo(`Waiting for realtime notifications on "${this.args.index}":"${this.args.collection}" ...`)
+
+    // eslint-disable-next-line @typescript-eslint/no-empty-function
+    await new Promise(() => { })
```
rolljee, line 97: Could this be juste Promise.resolve(); ?

src/commands/sdk/query.ts
Aschen: It's because the size of imported code is smaller?

src/commands/subscribe.ts
Aschen: Precisely I don't want the promise to resolve, so the command will wait indefinitely because some realtime subscription is still running

src/commands/sdk/query.ts
Aschen: In some files, I need to import many methods from lodash `isEmpty`, `isPlainObject`, etc
I don't like it much to have 5 lines of import instead of one :(

rolljee: The overall importe size is far lower indeed :)
Screenshot 2020-05-15 at 09 42 58

src/commands/subscribe.ts
rolljee: Syntax seemed odd to me, but i understand why 👍

src/commands/sdk/query.ts
Aschen: I agree but since this is not a web application people have to download lodash entirely so it won't change the download time.
But it's still a good point, I'm gonna add every lodash modules separately to reduce the total size

Aschen: EDIT: since `lodash` is already a dependency of Kourou, even if I use native methods we won't gain any size..

scottinet: Also, required modules are cached, so if the same module is required in multiple files, it doesn't change neither the memory consumption nor the time taken to require the code.

src/commands/subscribe.ts
scottinet: I don't think that this is needed. When subscribing, the SDK keeps the socket opened and will indefinitely wait for new messages, so the event loop is never empty. 
Until a kuzzle.disconnect() is called, which will never happen.

scottinet:
Or better yet: use [kill](https://nodejs.org/dist/latest-v12.x/docs/api/child_process.html#child_process_subprocess_kill_signal) on the spawned child process using a dedicated cucumber hook?

src/commands/subscribe.ts
Aschen: Yes but in fact a call to `kuzzle.disconnect` is made when the `run` method returns (see [here](https://github.com/kuzzleio/kourou/blob/master/src/common.ts#L75) )
This is why I need to wait indefinitely  here

src/commands/sdk/query.ts
rolljee: That's true server side indeed, usually a good practice client side :)

PR #44, opened 2020-05-13
Add user:export and user:import

Aschen:
## What does this PR do?

This PR adds two commands to export and import users.  
User are not exported with their credentials since Kuzzle can't access them.

It is still useful to export users and then we can either create credentials for them or ask them to recreate a password (thanks to the new password policies)

Also add the users import to the global `import` command

```
USAGE
  $ kourou user:import PATH

ARGUMENTS
  PATH  Dump file

OPTIONS
  --as=as              Impersonate a user
  --help               show CLI help
  --host=host          [default: localhost] Kuzzle server host
  --password=password  Kuzzle user password
  --port=port          [default: 7512] Kuzzle server port
  --protocol=protocol  [default: ws] Kuzzle protocol (http or websocket)
  --ssl                Use SSL to connect to Kuzzle
  --username=username  [default: anonymous] Kuzzle username (local strategy)
```

```
USAGE
  $ kourou user:export

OPTIONS
  --as=as              Impersonate a user
  --exclude=exclude    [default: []] Exclude user by matching their IDs
  --help               show CLI help
  --host=host          [default: localhost] Kuzzle server host
  --password=password  Kuzzle user password
  --path=path          [default: users] Dump directory
  --port=port          [default: 7512] Kuzzle server port
  --protocol=protocol  [default: ws] Kuzzle protocol (http or websocket)
  --ssl                Use SSL to connect to Kuzzle
  --username=username  [default: anonymous] Kuzzle username (local strategy)
```

Run the tests

### Other changes

 - fix waiting of `restoreUsers` promises
 - fix global import command 

### Boyscout

 - default value for refresh in the security controller is already `wait_for`

### DIFF
--- a/README.md
+++ b/README.md
@@ -897,7 +897,7 @@ USAGE
 
 OPTIONS
   --as=as              Impersonate a user
-  --exclude=exclude    [default: []] Exclude user by matching their IDs
+  --exclude=exclude    [default: []] Exclude users by matching their IDs
   --help               show CLI help
   --host=host          [default: localhost] Kuzzle server host
   --password=password  Kuzzle user password
--- a/README.md
+++ b/README.md
@@ -913,7 +913,7 @@ OPTIONS
 DESCRIPTION
   Exports users
 
-  Users credentials are NOT exported but only users content and profiles.
+  Users credentials are NOT exported. Only users' content and profiles are.
 
   See https://docs.kuzzle.io/core/2/guides/essentials/export-import-data.
 ```
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ $ npm install -g kourou
 $ kourou COMMAND
 running command...
 $ kourou (-v|--version|version)
-kourou/0.12.0 linux-x64 node-v12.16.0
+kourou/0.12.0 linux-x64 node-v12.16.2
 $ kourou --help [COMMAND]
 USAGE
   $ kourou COMMAND
@@ -909,6 +909,13 @@ OPTIONS
   --protocol=protocol      [default: ws] Kuzzle protocol (http or websocket)
   --ssl                    Use SSL to connect to Kuzzle
   --username=username      [default: anonymous] Kuzzle username (local strategy)
+
+DESCRIPTION
+  Exports users
+
+  Users credentials are NOT exported but only users content and profiles.
+
+  See https://docs.kuzzle.io/core/2/guides/essentials/export-import-data.
 ```
 
 _See code: [src/commands/user/export.ts](https://github.com/kuzzleio/kourou/blob/v0.12.0/src/commands/user/export.ts)_
--- a/README.md
+++ b/README.md
@@ -898,16 +898,17 @@ USAGE
   $ kourou user:export
 
 OPTIONS
-  --as=as              Impersonate a user
-  --exclude=exclude    [default: []] Exclude users by matching their IDs
-  --help               show CLI help
-  --host=host          [default: localhost] Kuzzle server host
-  --password=password  Kuzzle user password
-  --path=path          [default: users] Dump directory
-  --port=port          [default: 7512] Kuzzle server port
-  --protocol=protocol  [default: ws] Kuzzle protocol (http or websocket)
-  --ssl                Use SSL to connect to Kuzzle
-  --username=username  [default: anonymous] Kuzzle username (local strategy)
+  --as=as                  Impersonate a user
+  --batch-size=batch-size  [default: 2000] Maximum batch size (see limits.documentsFetchCount config)
+  --exclude=exclude        [default: []] Exclude users by matching their IDs
+  --help                   show CLI help
+  --host=host              [default: localhost] Kuzzle server host
+  --password=password      Kuzzle user password
+  --path=path              [default: users] Dump directory
+  --port=port              [default: 7512] Kuzzle server port
+  --protocol=protocol      [default: ws] Kuzzle protocol (http or websocket)
+  --ssl                    Use SSL to connect to Kuzzle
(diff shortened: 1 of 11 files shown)


src/commands/user/export.ts
```diff
@@ -0,0 +1,75 @@
...
+
+    this.logOk(`${Object.keys(users).length} users dumped`)
+  }
+
+  async _dumpUsers() {
+    const users: any = {}
+
+    let results = await this.sdk?.security.searchUsers(
+      {},
+      { scroll: '5s', size: 100 })
```
alexandrebouthinon, line 57: IMHO scroll duration and size should be configurable.

Aschen: Why? Kuzzle with automatically refresh the `scroll` between each call so those `5s` represent the maximal time to process the 100 users.
Even with a bigger batch, the only operation is to put the users in memory so `5s` is far than enough

About the size I will add the `batch-size` option in case people want to dump bigger or smaller batches.

This method may not work if you have something like 10K or 100K users since it keep all the users in memory before saving them. If we want to handle large amount of users we need to use stream like for the collection import/export but it will complexify the code a lot so IMHO we can stick with this for now

alexandrebouthinon: I don't really like magic numbers in the code of an application or a library. I think you should make them configurable but keep the values you chose by default. :slightly_smiling_face:

alexandrebouthinon: At least for the scroll size parameter :wink:

Aschen: Done :+1: And again about the scroll duration, IMHO it's not a magic number or whatever, I just need to put a duration to open a scroll session. In this use case there is no reason for the user to put a higher scroll duration and actually it might be dangerous to let users do so

src/support/restore-securities.ts
```diff
@@ -1,27 +1,73 @@
...
+  }
+
+  const anonymousRights = _.get(dump.content, 'anonymous.controllers.*.actions.*')
+
+  if (!preserveAnonymous && anonymousRights === false) {
+    if (command.sdk.username === 'anonymous') {
+      command.logKo('You are currently loggued as "anonymous" and anonymous role write will be overwritten.')
+      command.logInfo('Use the --preserve-anonymous flag to keep default anonymous rights.')
+
+      throw new Error('Please authenticate before importing or use --preserve-anonymous.')
```
scottinet, line 17: I don't get why this is a problem with anonymous, and not with any other rights. And I'm not sure that "restoring all roles apart from anonymous" is that useful, as anonymous is usually the first one to be (severely) restricted.

Wouldn't it be better to use `admin:loadSecurities` to prevent that kind of problem, as this route acts as an atomic way to restore roles, profiles and users?

Aschen: IMHO most of the time people use the export/import commands of the CLI to bring the production data into their local setup in order to reproduce a bug.

In production the anonymous user is restricted but in local it's common to keep the default rights.

When you are importing users, roles, profiles and collections with a single `import` command, after loading the anonymous rights every subsequent import will fail because the anonymous user is now restricted.

I feel this is very related to Kourou and the `import` command so that's why I put the logic here and not into Kuzzle.

scottinet: > Also, if you import the roles but you didn't create any user on your local instance then you are doomed.

Oh. Of course. There is that. :expressionless:

README.md
```diff
@@ -125,12 +127,12 @@ ARGUMENTS
...
-  --host=host          [default: localhost] Kuzzle server host
-  --password=password  Kuzzle user password
-  --port=port          [default: 7512] Kuzzle server port
+  --host=host          [default: api.v2.myomniscient.com] Kuzzle server host
+  --password=password  [default: 0szBh8W7dHMI/u2nQRr6RoZbj+KSlty3H6ySPUe8wmY=] Kuzzle user password
+  --port=port          [default: 443] Kuzzle server port
   --protocol=protocol  [default: http] Kuzzle protocol (http or ws)
   --ssl                Use SSL to connect to Kuzzle
-  --username=username  [default: anonymous] Kuzzle username (local strategy)
+  --username=username  [default: admin-kuzzle-omniscient] Kuzzle username (local strategy)
```
scottinet, line 135: ... is that what I think this is? :astonished:

README.md
```diff
@@ -885,6 +889,61 @@ DESCRIPTION
...
+  --path=path              [default: users] Dump directory
+  --port=port              [default: 443] Kuzzle server port
+  --protocol=protocol      [default: ws] Kuzzle protocol (http or websocket)
+  --ssl                    Use SSL to connect to Kuzzle
+  --username=username      [default: admin-kuzzle-omniscient] Kuzzle username (local strategy)
+
+DESCRIPTION
+  Exports users
+
+  Users credentials are NOT exported but only users content and profiles.
```
scottinet, line 916:
```suggestion
  Users credentials are NOT exported. Only users' content and profiles are.
```

README.md
```diff
@@ -125,12 +127,12 @@ ARGUMENTS
 OPTIONS
   --as=as              Impersonate a user
   --help               show CLI help
-  --host=host          [default: localhost] Kuzzle server host
-  --password=password  Kuzzle user password
-  --port=port          [default: 7512] Kuzzle server port
+  --host=host          [default: api.v2.myomniscient.com] Kuzzle server host
```
scottinet, line 130: ... is that what I think this is? :astonished:

src/commands/user/export.ts
scottinet: A bit less dangerous now that scroll contexts are automatically destroyed by Kuzzle once you fetch all its results ;)

PR #51, opened 2020-06-19
[fix] readme links should be relative

Leodau:
## What does this PR do?
Updates all _readme_ file links to be relative instead of absolute by using oclif's **[repositoryPrefix](https://github.com/oclif/dev-cli/blob/v1.22.2/src/commands/readme.ts#L170)**

1. Checkout to `fix/readme-links` branch
2. Click on any `'see code'` command link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,9 +64,9 @@ User impersonation require the following rights for the authenticated user: `sec
 
 ```bash
 $ kourou sdk:query auth:getCurrentUser --as gordon --username admin --password admin
- 
+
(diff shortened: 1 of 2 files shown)


Aschen:
The README is autogenerated by ocliff so it will be overwritten :/

Leodau:
> The README is autogenerated by ocliff so it will be overwritten :/

Oh i didn't know, i'll check if we could change that on the ocliff side then

PR #54, opened 2020-06-24
instance:list and instance:kill commands

berthieresteban:
## What does this PR do?
First contribution to kourou and first use of ts, please be lenient.

Add the `kourou instance:list` and the `kourou instance:kill` commands

  - Step 1 : Run some kuzzle instance from both 1 & 2 versions
  - Step 2 : `kourou instance:list` should display informations about your current running stacks
  - Step 3 : `kourou instance:kill -i stack-0_kuzzle_1` should kill the containers of the stack 0
  - Step 4 : `kourou instance:kill` should display a prompt to select the instance to kill
  - Step 5 : `kourou instance:kill --all` should kill all running instances
 

### Other changes
/

### Boyscout
Throw if prompt must appear using `kourou instance:logs` to select a kuzzle instance but no one is running
Tests for `instance:spawn`

### Screenshots
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ $ npm install -g kourou
 $ kourou COMMAND
 running command...
 $ kourou (-v|--version|version)
-kourou/0.13.0 linux-x64 node-v10.15.0
+kourou/0.13.0 linux-x64 node-v12.16.3
 $ kourou --help [COMMAND]
 USAGE
   $ kourou COMMAND
--- a/README.md
+++ b/README.md
@@ -64,9 +64,9 @@ User impersonation require the following rights for the authenticated user: `sec
 
 ```bash
 $ kourou sdk:query auth:getCurrentUser --as gordon --username admin --password admin
- 
+
  🚀 Kourou - Executes an API query.
- 
+
  [ℹ] Connecting to http://localhost:7512 ...
  [ℹ] Impersonate user "gordon"
 
--- a/README.md
+++ b/README.md
@@ -24,7 +24,7 @@ $ npm install -g kourou
 $ kourou COMMAND
 running command...
 $ kourou (-v|--version|version)
-kourou/0.13.0 linux-x64 node-v12.16.3
+kourou/0.13.0 linux-x64 node-v10.15.0
 $ kourou --help [COMMAND]
 USAGE
   $ kourou COMMAND
@@ -97,6 +97,8 @@ $ kourou sdk:query auth:getCurrentUser --as gordon --username admin --password a
 * [`kourou import PATH`](#kourou-import-path)
 * [`kourou index:export INDEX`](#kourou-indexexport-index)
 * [`kourou index:import PATH`](#kourou-indeximport-path)
+* [`kourou instance:kill`](#kourou-instancekill)
+* [`kourou instance:list`](#kourou-instancelist)
 * [`kourou instance:logs`](#kourou-instancelogs)
 * [`kourou instance:spawn`](#kourou-instancespawn)
 * [`kourou profile:export`](#kourou-profileexport)
@@ -656,6 +658,31 @@ EXAMPLES
 
 _See code: [src/commands/index/import.ts](https://github.com/kuzzleio/kourou/blob/v0.13.0/src/commands/index/import.ts)_
 
+## `kourou instance:kill`
+
+Kill all the containers of a running kuzzle instance
+
+```
+USAGE
+  $ kourou instance:kill
+
+OPTIONS
+  -i, --instance=instance  Kuzzle instance name
+```
+
+_See code: [src/commands/instance/kill.ts](https://github.com/kuzzleio/kourou/blob/v0.13.0/src/commands/instance/kill.ts)_
+
+## `kourou instance:list`
+
+Lists the Kuzzle running instances
+
+```
+USAGE
+  $ kourou instance:list
+```
+
+_See code: [src/commands/instance/list.ts](https://github.com/kuzzleio/kourou/blob/v0.13.0/src/commands/instance/list.ts)_
+
 ## `kourou instance:logs`
 
 Displays the logs of a running Kuzzle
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -24,7 +24,7 @@ $ npm install -g kourou
 $ kourou COMMAND
 running command...
 $ kourou (-v|--version|version)
-kourou/0.13.0 linux-x64 node-v12.16.3
+kourou/0.13.0 linux-x64 node-v10.15.0
```
Aschen, line 27: You should use Node.js 12 by default

src/commands/instance/kill.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  static flags = {
+    instance: flags.string({
+      char: 'i',
+      description: 'Kuzzle instance name'
+    })
+  }
+
+  async runSafe() {
+    let instance: string = this.flags.instance!
```
Aschen, line 20: What does the `!` do at the end?

src/commands/instance/kill.ts
```diff
@@ -0,0 +1,83 @@
...
+    const esName = `${stack}_elasticsearch_1`
+    const redisName = `${stack}_redis_1`
+    const args = ['kill', kuzzleName, esName, redisName]
+
+    const instanceKill = execa('docker', args)
+
+    instanceKill.stdout.pipe(process.stdout)
+    instanceKill.stderr.pipe(process.stderr)
+
+    await instanceKill
```
Aschen, line 53: Useless await, you can directly returns the promise

src/commands/instance/kill.ts
```diff
@@ -0,0 +1,83 @@
...
+  private async getInstancesList(): Promise<string[]> {
+    let containersListProcess
+
+    try {
+      containersListProcess = await execa('docker', [
+        'ps',
+        '--format',
+        '"{{.Names}}"'
+      ])
+    } catch {
```
Aschen, line 65: Nit: this `catch` should be on the next line

src/commands/instance/kill.ts
```diff
@@ -0,0 +1,83 @@
...
+      this.warn(
```
Aschen, line 66: Nit: this should be on the same line

src/commands/instance/kill.ts
```diff
@@ -0,0 +1,83 @@
...
+      return []
+    }
+
+    const containersList: string[] = containersListProcess.stdout
+      .replace(/"/g, '')
+      .split('\n')
+
+    return containersList.filter(
+      containerName =>
+        containerName.includes('kuzzle') &&
```
Aschen, line 78: Nit: operators should be on the beginning of the line

src/commands/instance/list.ts
```diff
@@ -0,0 +1,106 @@
...
+  private async getInstancesList(): Promise<object[]> {
+    let containersListProcess
+
+    try {
+      containersListProcess = await execa('docker', [
+        'ps',
+        '--format',
+        '"{{.Names}}%{{.Image}}%{{.Status}}%{{.Ports}}"'
+      ])
+    } catch {
```
Aschen, line 36: Ditto

src/commands/instance/list.ts
```diff
@@ -0,0 +1,106 @@
...
+      this.warn(
```
Aschen, line 37: Ditto

src/commands/instance/list.ts
```diff
@@ -0,0 +1,106 @@
...
+      esPort: 0
+    }))
+
+    for (const container of containersList) {
+      const splitted: string[] = container.split('%')
+
+      const stackNumber: number = parseInt(
+        splitted[0].split('_')[0].split('-')[1],
+        10
+      )
```
Aschen, line 68: Nit: this should be on the same line :top:

src/commands/instance/kill.ts
berthieresteban: https://www.logicbig.com/tutorials/misc/typescript/non-null-assertion-operator.html

Aschen: You should remove it, it's useless since you can't ensure that this flag is not null. It's only set (or not) at runtime

src/commands/instance/list.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export class InstanceList extends Kommand {
+  static initSdk = false
+
+  static description = 'Lists the Kuzzle running instances'
+
+  async runSafe() {
+    const instancesList = await this.getInstancesList()
+    // Disable eslint no-console to allow using console.table
+    // Instead of add another dependency to kourou
```
Leodau, line 13: Console.table is self explanatory :100: 
```suggestion
    // Disable eslint no-console to allow using console.table
    // Instead of add another dependency to kourou
```

src/commands/instance/list.ts
```diff
@@ -0,0 +1,79 @@
...
+          formatedStacks[stackNumber].kuzzlePort = parseInt(port.substring(port.indexOf('->7512/tcp') - 4, port.indexOf('->7512/tcp')), 10)
+          break
+        case 'redis':
+          formatedStacks[stackNumber].redisVersion = version
+          formatedStacks[stackNumber].redisPort = parseInt(port.replace('/tcp', ''), 10)
+          break
+        case 'elasticsearch':
+          formatedStacks[stackNumber].esVersion = version
+          formatedStacks[stackNumber].esPort = parseInt(port.substring(port.indexOf('->9200/tcp') - 4, port.indexOf('->9200/tcp')), 10)
+          break
```
Leodau, line 73: What if some other unknown stacks are running? What if the services ports are customized? (if that is even possible)

src/commands/instance/logs.ts
```diff
@@ -21,11 +21,14 @@ export class InstanceLogs extends Kommand {
...
 
   async runSafe() {
-    let instance: string = this.flags.instance!
+    let instance: string = this.flags.instance
     const followOption: boolean = this.flags.follow
 
     if (!instance) {
       const instancesList = await this.getInstancesList()
+      if (instancesList.length === 0) {
+        throw new Error('There is no Kuzzle running instances')
```
Leodau, line 30:
```suggestion
        throw new Error('There are no Kuzzle running instances')
```

src/commands/instance/list.ts
berthieresteban: That `kourou instance:list` command only work for instances spawned using kourou.
I'll add a regex test to be sure we only use the containers matching the kourou syntax (`stack-x_kuzzle_1`)
I'm not sure we are able to check instances spawned directly by the user since he can name the containers like he wants
Concerning the ports, since we cannot customize them using `kourou instance:spawn` that should not be a problem

src/commands/instance/kill.ts
```diff
@@ -14,25 +14,39 @@ export class InstanceLogs extends Kommand {
...
+
+    if (!instance) {
       const responses: any = await inquirer.prompt([{
           name: 'instance',
           message: 'Which kuzzle instance do you want to kill',
           type: 'list',
           choices: instancesList
       }])
       instance = responses.instance!
+    } else if (!instancesList.includes(instance)) {
```
Aschen, line 47: Nitpicking: should be on the next line

PR #55, opened 2020-06-24
User mapping import export

berthieresteban:
## What does this PR do?

add `user:exportMapping` and `user:importMapping` commands using `security.getUserMapping` and `security.updateUserMapping`

  - Step 1 : `kourou instance:spawn`
  - Step 2 : `kourou user:exportMapping`
  - Step 3 :  check your `user/userMapping.json` file and update it by adding some fields next to `profileIds` property.
  - Step 4 :  `kourou user:importMapping user/userMapping.json` 
  - Step 5 : Check on next-console.kuzzle.io your user mapping in security tab
 

### Other changes
/

### Boyscout
/


src/commands/import.ts
```diff
@@ -87,8 +107,24 @@ export default class Import extends Kommand {
         const total = await restoreProfiles(this, dump)
 
         this.logOk(`[profiles] Imported ${total} profiles`)
+      } catch (error) {
+        this.logKo(`Error during import of ${file}: ${error.message}. Skipped.`)
       }
-      catch (error) {
+    }
+    if (files.userMapping.length !== 0) {
```
Aschen, line 114: You should follow the same logic with the  `for` loop. Even if there is little chances that someone import 2 users mappings, it could still happen.

src/commands/import.ts
```diff
@@ -87,8 +107,24 @@ export default class Import extends Kommand {
...
+      const file = files.userMapping[0]
+      try {
+        this.logInfo(`[user mapping] Start importing user mapping in ${file}`)
+        const dump = JSON.parse(fs.readFileSync(file, 'utf8'))
+
+        const mapping = dump.content.mapping
+        delete mapping.profileIds
```
Aschen, line 121: Why deleting `profileIds`? They are part of the mappings (even if I agree that yes they should already exists still they are part of the default mappings)

src/commands/user/export-mapping.ts
```diff
@@ -0,0 +1,42 @@
+import fs from 'fs'
+import path from 'path'
+
+import { flags } from '@oclif/command'
+import { Kommand } from '../../common'
+import { kuzzleFlags } from '../../support/kuzzle'
+
+export default class UserExportMapping extends Kommand {
+  static description = 'Exports user mapping to JSON.'
```
Aschen, line 9: Mappings should always be used in the plural form (I know the API route is `updateMapping` but that's an error..)
```suggestion
  static description = 'Exports user mappings to JSON.'
```

src/commands/user/export-mapping.ts
```diff
@@ -0,0 +1,42 @@
...
+
+  static flags = {
+    help: flags.help({}),
+    path: flags.string({
+      description: 'Dump directory',
+      default: 'userMapping'
```
Aschen, line 15: Should be the same as users data (collection data and mappings are dump in the same folder)
```suggestion
      default: 'users'
```

src/commands/user/export-mapping.ts
```diff
@@ -0,0 +1,42 @@
...
+    }),
+    ...kuzzleFlags,
+    protocol: flags.string({
+      description: 'Kuzzle protocol (http or websocket)',
+      default: 'ws'
+    })
+  }
+
+  async runSafe() {
+    const filename = path.join(this.flags.path, 'userMapping.json')
```
Aschen, line 25: Mappings should be in the plural form, also we are talking about the `users` collection
```suggestion
    const filename = path.join(this.flags.path, 'users-mappings.json')
```

src/commands/user/export-mapping.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    this.logInfo(`Exporting user mapping in ${filename} ...`)
```
Aschen, line 27:
```suggestion
    this.logInfo(`Exporting users mappings in ${filename} ...`)
```

src/commands/user/export-mapping.ts
```diff
@@ -0,0 +1,42 @@
...
+    const mapping = await this.sdk?.security.getUserMapping()
+
+    const dump = {
+      type: 'userMapping',
+      content: mapping
+    }
+
+    fs.writeFileSync(filename, JSON.stringify(dump, null, 2))
+
+    this.logOk('User mapping dumped')
```
Aschen, line 40:
```suggestion
    this.logOk('Users mappings dumped')
```

src/commands/user/export-mapping.ts
```diff
@@ -0,0 +1,42 @@
...
+    const filename = path.join(this.flags.path, 'userMapping.json')
+
+    this.logInfo(`Exporting user mapping in ${filename} ...`)
+
+    fs.mkdirSync(this.flags.path, { recursive: true })
+
+    const mapping = await this.sdk?.security.getUserMapping()
+
+    const dump = {
+      type: 'userMapping',
```
Aschen, line 34: About the kebab-case notation, IMHO it's more readable because it's closer to the natural sentence construction (word separated by spaces). Since variable names cannot contains `-`, we use camelCase (or snake_case which I prefer for the same reason)
```suggestion
      type: 'users-mappings',
```

src/commands/user/import-mapping.ts
```diff
@@ -0,0 +1,34 @@
+import { flags } from '@oclif/command'
+import fs from 'fs'
+
+import { Kommand } from '../../common'
+import { kuzzleFlags } from '../../support/kuzzle'
+
+export default class UserImport extends Kommand {
+  static description = 'Imports users'
```
Aschen, line 8:
```suggestion
  static description = 'Imports users mappings'
```

src/commands/user/import-mapping.ts
```diff
@@ -0,0 +1,34 @@
+import { flags } from '@oclif/command'
+import fs from 'fs'
+
+import { Kommand } from '../../common'
+import { kuzzleFlags } from '../../support/kuzzle'
+
+export default class UserImport extends Kommand {
```
Aschen, line 7:
```suggestion
export default class UserImportMappings extends Kommand {
```

src/commands/user/import-mapping.ts
```diff
@@ -0,0 +1,34 @@
...
+    protocol: flags.string({
+      description: 'Kuzzle protocol (http or websocket)',
+      default: 'ws'
+    })
+  }
+
+  static args = [{ name: 'path', description: 'Dump file', required: true }]
+
+  async runSafe() {
+    this.logInfo(`Importing user mapping from ${this.args.path} ...`)
```
Aschen, line 22:
```suggestion
    this.logInfo(`Importing users mappings from ${this.args.path} ...`)
```

src/commands/user/import-mapping.ts
```diff
@@ -0,0 +1,34 @@
...
+
+    const dump = JSON.parse(fs.readFileSync(this.args.path, 'utf-8'))
+
+    const mapping = dump.content.mapping
+    delete mapping.profileIds
+
+    await this.sdk?.security.updateUserMapping({
+      properties: mapping
+    })
+    this.logOk('User mapping restored')
```
Aschen, line 32:
```suggestion
    this.logOk('Users mappings restored')
```

src/commands/import.ts
berthieresteban: Not sure of the comportement needed here.
If someone has two users mappings it means that he want both to be imported ? 
We might throw or skip the mapping import and display something like `There are multiple users mappings, you can import the one you want using following command kourou users:importMapping {filepath}`

src/commands/import.ts
berthieresteban: I just followed the example of https://docs.kuzzle.io/sdk/js/7/controllers/security/update-user-mapping/ wich doesn't add profileIds file in the body.
Should i let this field ?

src/commands/user/export-mapping.ts
berthieresteban: `userS mappings` ?

src/commands/import.ts
Aschen: Multiple users mappings can be imported from different files as long as they don't define the same field twice but with a different type.

But seriously I doubt someone will do this actually :sweat_smile: This comment was mostly to have a uniform code

src/commands/import.ts
Aschen: I would say yes, this is harmless and also IMHO it's a good thing to not hide this from users

features/step_definitions/common/security-steps.js
```diff
@@ -111,3 +112,22 @@ Given('The user {string} should match:', async function (userId, dataTable) {
 
   should(user.content).match(expectedContent);
 })
+
+Given('I create an {string} file with content:', async function (filename, dataTable) {
```
Aschen, line 116: The step name should be more specific since it's specific to `userMappings` dump files

features/User.feature
```diff
@@ -25,4 +25,22 @@ Feature: User Commands
       | profileIds | ["admin"]           |
       | email      | "alyx@blackmesa.us" |
 
+  # user:export-mapping & user:import-mapping ==================================================
 
+  @security
+  Scenario: Export and import users mappings
```
Leodau, line 31: No need to add again the plural form, user mappings; can mean multiple users and/or multiple mappings.
```suggestion
  Scenario: Export and import user mappings
```

features/User.feature
```diff
@@ -25,4 +25,22 @@ Feature: User Commands
...
+      | arg | ./dump/user-mapping.json |
+    Then I successfully call the route "security":"getUserMapping"
+    And The property "mapping" of the result should match:
+      | profileIds | {"type": "keyword"} |
+      | email | { "type": "keyword" }    |
+      | age   | { "type": "integer" }     |
+    When I run the command "user:export-mappings" with:
+      | flag | --path    | ./dump     |
+    Then The file "./dump/users-mappings.json" content should match:
+      | type    | "usersMappings"                                                                                                    |
```
Leodau, line 45: Why the kommand **user:export-mappings** would generate a file named user**s**-mappings.json. I do even feel like, we don't export multiple mappings, we export A mapping. Is there multiple mappings?

features/step_definitions/common/security-steps.js
```diff
@@ -111,3 +112,22 @@ Given('The user {string} should match:', async function (userId, dataTable) {
...
+  const content = this.parseObject(dataTable)
+
+  const body = {
+    type: 'usersMappings',
```
Leodau, line 120: Again, is this correct? is there a different mapping for **each** user?

src/commands/import.ts
```diff
@@ -92,6 +92,20 @@ export default class Import extends Kommand {
...
     }
+    for (const file of files.usersMappings) {
+      try {
+        this.logInfo(`[users] Start importing users mappings in ${file}`)
+        const dump = JSON.parse(fs.readFileSync(file, 'utf8'))
+
+        const mapping = dump.content.mapping
+        await this.sdk?.security.updateUserMapping({ properties: mapping })
+
+        this.logOk('[users] mappings imported')
```
Leodau, line 103: Again, why would this be in plural. Are there any differences with these:
https://github.com/kuzzleio/kourou/blob/e180dd491540502315f642c3c0586140b3324901/src/commands/import.ts#L61
https://github.com/kuzzleio/kourou/blob/c3695994abfd264ae05cc2143ca8770bf7e3c41d/src/commands/document/create.ts#L57

src/commands/import.ts
```diff
@@ -124,15 +138,16 @@ export default class Import extends Kommand {
         memo[fileType].push(file)
 
         return memo
-      }, { documents: [], mappings: [], roles: [], profiles: [], users: [] })
+      }, { documents: [], mappings: [], roles: [], profiles: [], usersMappings: [], users: [] })
```
Leodau, line 141:
```suggestion
      }, { documents: [], mappings: [], roles: [], profiles: [], userMappings: [], users: [] })
```

src/commands/import.ts
```diff
@@ -124,15 +138,16 @@ export default class Import extends Kommand {
...
 
     for (const dir of directories) {
-      const { documents, mappings, roles, profiles, users } = await this.walkDirectories(dir)
+      const { documents, mappings, roles, profiles, usersMappings, users } = await this.walkDirectories(dir)
```
Leodau, line 144:
```suggestion
      const { documents, mappings, roles, profiles, userMappings, users } = await this.walkDirectories(dir)
```

src/commands/import.ts
```diff
@@ -124,15 +138,16 @@ export default class Import extends Kommand {
...
 
       files.documents = files.documents.concat(documents)
       files.mappings = files.mappings.concat(mappings)
       files.roles = files.roles.concat(roles)
       files.profiles = files.profiles.concat(profiles)
+      files.usersMappings = files.usersMappings.concat(usersMappings)
```
Leodau, line 150:
```suggestion
      files.userMappings = files.usersMappings.concat(usersMappings)
```

src/commands/user/export-mappings.ts
```diff
@@ -0,0 +1,42 @@
+import fs from 'fs'
+import path from 'path'
+
+import { flags } from '@oclif/command'
+import { Kommand } from '../../common'
+import { kuzzleFlags } from '../../support/kuzzle'
+
+export default class UserExportMappings extends Kommand {
+  static description = 'Exports users mappings to JSON.'
```
Leodau, line 9:
```suggestion
  static description = 'Exports user mappings to JSON.'
```

src/commands/user/export-mappings.ts
```diff
@@ -0,0 +1,42 @@
...
+    protocol: flags.string({
+      description: 'Kuzzle protocol (http or websocket)',
+      default: 'ws'
+    })
+  }
+
+  async runSafe() {
+    const filename = path.join(this.flags.path, 'users-mappings.json')
+
+    this.logInfo(`Exporting users mappings in ${filename} ...`)
```
Leodau, line 27:
```suggestion
    this.logInfo(`Exporting user mappings in ${filename} ...`)
```

src/commands/user/export-mappings.ts
```diff
@@ -0,0 +1,42 @@
...
+    }),
+    ...kuzzleFlags,
+    protocol: flags.string({
+      description: 'Kuzzle protocol (http or websocket)',
+      default: 'ws'
+    })
+  }
+
+  async runSafe() {
+    const filename = path.join(this.flags.path, 'users-mappings.json')
```
Leodau, line 25:
```suggestion
    const filename = path.join(this.flags.path, 'user-mappings.json')
```

src/commands/user/export-mappings.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    this.logInfo(`Exporting users mappings in ${filename} ...`)
+
+    fs.mkdirSync(this.flags.path, { recursive: true })
+
+    const mapping = await this.sdk?.security.getUserMapping()
+
+    const dump = {
+      type: 'usersMappings',
```
Leodau, line 34:
```suggestion
      type: 'userMappings',
```

src/commands/user/export-mappings.ts
```diff
@@ -0,0 +1,42 @@
...
+      content: mapping
+    }
+
+    fs.writeFileSync(filename, JSON.stringify(dump, null, 2))
+
+    this.logOk('Users mappings dumped')
```
Leodau, line 40: If we actually generate a different mapping PER user, then it should be something like: User's mappings. or Every user mapping.
```suggestion
    this.logOk('User mappings dumped')
```

src/commands/user/import-mappings.ts
```diff
@@ -0,0 +1,31 @@
+import { flags } from '@oclif/command'
+import fs from 'fs'
+
+import { Kommand } from '../../common'
+import { kuzzleFlags } from '../../support/kuzzle'
+
+export default class UserImportMappings extends Kommand {
+  static description = 'Imports users mappings'
```
Leodau, line 8:
```suggestion
  static description = 'Imports user mappings'
```

src/commands/user/import-mappings.ts
```diff
@@ -0,0 +1,31 @@
...
+    protocol: flags.string({
+      description: 'Kuzzle protocol (http or websocket)',
+      default: 'ws'
+    })
+  }
+
+  static args = [{ name: 'path', description: 'Dump file', required: true }]
+
+  async runSafe() {
+    this.logInfo(`Importing users mappings from ${this.args.path} ...`)
```
Leodau, line 22:
```suggestion
    this.logInfo(`Importing user mappings from ${this.args.path} ...`)
```

src/commands/user/import-mappings.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    const dump = JSON.parse(fs.readFileSync(this.args.path, 'utf-8'))
+
+    const mapping = dump.content.mapping
+
+    await this.sdk?.security.updateUserMapping({ properties: mapping })
+    this.logOk('Users mappings restored')
```
Leodau, line 29:
```suggestion
    this.logOk('User mappings restored')
```

Aschen:
> I'm not sure about the usage of the plural form of **users**, is there really multiple mappings PER user, and multiple users in the mappings? (cuz that's the only way we can say that they are indeed **users mappings**)

We are talking about a collection named `users` and not about a user document, that's why we use the plural form (plural form for collection, singular for documents)

PR #62, opened 2020-09-30
Adds app:scaffold and app:run commands

Aschen:
## What does this PR do?

Adds the `kourou app:scaffold <app name>` and `kourou app:run` commands.

The `app:scaffold` command creates a new directory and copy the `package.json` and the `app.ts` files into it. Then it install the latest Kuzzle version and the development dependencies needed for Typescript

It will creates the following structure for the new project
image

## Other changes

 - remove `execa` dependency
 - support windows for path


src/commands/app/run.ts
```diff
@@ -0,0 +1,24 @@
...
+  static description = 'Run the Kuzzle application in the current directory'
+
+  static flags = {
+    help: flags.help(),
+  }
+
+  async runSafe() {
+    await AppStartServices.run()
+
+    await execute('node', '-r', 'ts-node/register', 'app.ts')
```
xbill82, line 22: What's the benefit of this command versus putting an `npm run dev` script (running the same command) into the app scaffold?

src/commands/app/scaffold.ts
```diff
@@ -0,0 +1,66 @@
...
+    help: flags.help(),
+  }
+
+  static args = [
+    { name: 'name', description: 'Application name', required: true },
+  ]
+
+  async runSafe() {
+    this.logInfo(`Scaffold a new Kuzzle application in ${this.args.name}/`)
+    await execute('rm', '-rf', this.args.name)
```
xbill82, line 27: U sure about this? Wouldn't it be better to throw an error and let the user delete the directory?

src/support/execute.ts
```diff
@@ -0,0 +1,26 @@
+import { spawn } from 'child_process'
+
+export async function execute(...args: any[]): Promise<any> {
```
xbill82, line 3: Aren't we re-inventing [execa](https://www.npmjs.com/package/execa) here?

src/commands/app/scaffold.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    await execute('mkdir', this.args.name)
+
+    const templatePath = path.join(templatesDir, 'app-scaffold', 'ts')
+    await this.renderTemplates(templatePath, templatePath)
+
+    this.logInfo('Installing latest Kuzzle version via NPM...')
+
+    // await execute('npm', 'install', 'kuzzle', { cwd: this.args.name })
+    // await execute('npm', 'install', { cwd: this.args.name })
```
rolljee, line 36: Commented code should be removed

src/support/execute.ts
Aschen: I took in the other way, I removed `execa` :-)

xbill82: Why?

Aschen: To reduce the number of dependencies installed with the CLI.

scottinet (approved):

xbill82 (approved):

PR #67, opened 2020-10-14
Fix document:search command

Aschen:
## What does this PR do?

With the query hook, users are used to execute API action like this:
```
$ kourou <controller>:<action> <index> <collection> <body>
```

So I modified the `document:search` command be used with `kourou document:search <index> <collection> <query>`

### Other changes

 - add `fingerprint` to `api-key:search` command display
 - avoid `this.sdk?.<controller>` usage


scottinet:
shouldn't we simply remove the `document:search` command, then? :thinking:

Aschen:
The `document:search` command is still a "real" command because there is a few advantages to control the behavior:
 - better parameters (`--from 10` instead of `-a from=10`)
 - better body: only the query without specifying `query` inside the body, the sort is passed into a flag (`--sort`)
 - real documentation of the command
 - better display of the documents

scottinet:
It seems that only how arguments are provided makes the difference.  And I find that confusing. How, as a user, can I know in advance which one is executed?

I believe that we need to make additional commands stand alone more.

Aschen:
> It seems that only how arguments are provided makes the difference. And I find that confusing. How, as a user, can I know in advance which one is executed, between `document:search` and... well... `document:search`?
> 
> I believe that we need to make additional commands stand alone more.

The hook only works for non-existing commands. If kourou doesn't know the command, then it try to pass it to `sdk:query`.

So the `document:search` command is always executed with the original command and never with the inhered one

scottinet:
I guess we'll discuss this in the pre-release workshop anyway. :thinking:

scottinet (approved):

xbill82 (approved):

PR #76, opened 2020-10-30
Add snapshots support

### LINKED ISSUE #70: Add es:snapshot module
- [ ] `es:snapshot:dump <es-host> <repo_name> <snapshot_name>`
* Dump the ES data into a snapshot

- [ ] `es:snapshot:restore <es-host> <repo_name> <snapshot_name>`
* Restore a specific snapshot from a specific repository to an ES

- [ ] `es:snapshot:create-repository <es-host> <repo_name> <path_of_the_repo> --compress`
* Create a repository into an ES cluster

- [ ]  `es:snapshot:list-repository <es-host>`
* List repository known by the current ES

- [ ] `es:snapshot:list-snapshot <es-host> <repo_name>`
* List snapshot from a repository known by the current ES

<details><summary>Original issue</summary>
- [ ] `es:snapshot:dump --host=<es-host> --repository=<repo_name> --name=<snapshot_name>`
* Dump the ES data into a snapshot

- [ ] `es:snapshot:restore --host=<es-host> --repository=<repo_name> --snapshot=<snapshot_name>`
* Restore a specific snapshot from a specific repository to an ES

- [ ] `es:snapshot:create_repository --host=<es-host> --repository=<repo_name> --location=<path_of_the_repo> --compress=<true|false>`
* Create a repository into an ES cluster

- [ ]  `es:snapshot:list_repository --host=<es-host>`
* List repository known by the current ES

- [ ] `es:snapshot:list_snapshot --host=<es-host> --repository=<repo_name>`
* List snapshot from a repository known by the current ES
</details>

rolljee:
## What does this PR do?

This PR adds the snapshots module to kourou.

Close #70 

1. `git clone`
1. `npm install`
1. `./bin/run es:snapshots:<command> <args> --<flags>`


features/docker/elasticsearch/elasticsearch.yml
```diff
@@ -0,0 +1,5 @@
+cluster.name: "docker-cluster"
+network.host: 0.0.0.0
+
+# You need to create this folder in order to run snapshots
+path.repo: ["/tmp/snapshots"]
```
Aschen, line 5: Can't we pass this variable to ES with the environment instead of having to build a whole new Docker image?

Aschen (changes requested):

features/docker/elasticsearch/elasticsearch.yml
rolljee: I tried, does not work

Aschen: Why :thinking:

Aschen (approved):

src/commands/es/snapshots/restore.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  static args = [
+    { name: 'repository', description: 'ES repository name', required: true },
+    { name: 'name', description: 'ES snapshot name', required: true },
+  ]
+
+  async getIndices(esClient: Client) {
+    const beforeResponse = await esClient.cat.indices({ format: 'json' })
+
+    const indices: string[] = beforeResponse.body
```
Aschen, line 33: Type is automatically inferred here

```suggestion
    const indices = beforeResponse.body
```

src/commands/es/snapshots/restore.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    const esRequest = {
+      repository: this.args.repository,
+      snapshot: this.args.name,
+      body: {}
+    }
+
+    const response = await esClient.snapshot.restore(esRequest)
+
+    const indicesAfterRestore = await this.getIndices(esClient)
```
Aschen, line 58: You should call the method only one time and store the result

src/commands/es/snapshots/restore.ts
```diff
@@ -0,0 +1,65 @@
...
+    for (const index of indicesAfterRestore) {
+      await esClient.indices.open({ index })
+    }
+
+    this.logOk(`Success ${JSON.stringify(response.body)}`)
```
Aschen, line 63: I'm not sure about this. Can't we have a better message or at least an indented JSON?

tsconfig.json
```diff
@@ -4,15 +4,16 @@
     "importHelpers": true,
     "module": "commonjs",
     "outDir": "lib",
-    "rootDir": "src",
+    "rootDir": "./src",
     "strict": true,
     "target": "es2017",
-    "esModuleInterop": true
+    "esModuleInterop": true,
+    "resolveJsonModule": true
```
Aschen, line 11: I think we don't need this line anymore

src/commands/es/snapshots/dump.ts
```diff
@@ -0,0 +1,46 @@
+import { flags } from '@oclif/command'
+import { Client } from '@elastic/elasticsearch'
+
+import { Kommand } from '../../../common'
+
+export default class ESDump extends Kommand {
```
Aschen, line 6: Nitpicking
```suggestion
export default class EsSnapshotsDump extends Kommand {
```

src/commands/es/snapshots/create-repository.ts
```diff
@@ -0,0 +1,56 @@
+import { flags } from '@oclif/command'
+import { Client } from '@elastic/elasticsearch'
+
+import { Kommand } from '../../../common'
+
+export default class EsCreateSnapshot extends Kommand {
```
Aschen, line 6: Nitpicking
```suggestion
export default class EsSnapshotsCreateRepository extends Kommand {
```

src/commands/es/snapshots/list.ts
```diff
@@ -0,0 +1,59 @@
+import { flags } from '@oclif/command'
+import { Client } from '@elastic/elasticsearch'
+
+import { Kommand } from '../../../common'
+
+export default class ESlistSnapshot extends Kommand {
```
Aschen, line 6: Nitpicking
```suggestion
export default class EsSnapshotsList extends Kommand {
```

Aschen (changes requested):

src/commands/es/snapshots/restore.ts
rolljee: You are right, i did that because i wasn't sure indices where put in open state after a restore leading to closed indice. but I checked in the doc


`The restore operation must be performed on a functioning cluster. However, an existing index can be only restored if it’s closed and has the same number of shards as the index in the snapshot. The restore operation automatically opens restored indices if they were closed and creates new indices if they didn’t exist in the cluster.`

So we can safely remove this call indeed.

src/commands/es/snapshots/restore.ts
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    const esRequest = {
+      repository: this.args.repository,
+      snapshot: this.args.name,
+      body: {}
+    }
+
+    const response = await esClient.snapshot.restore(esRequest)
+
```
Aschen, line 57: You need to open the indices again and you should do that in a `finally` block to ensure they don't stay closed

Aschen (changes requested):

features/docker/elasticsearch/elasticsearch.yml
alexandrebouthinon: I think this is due to the fact that the value is an array. Could you try to use YAML objects for environment variables? Like this:
```yaml
services:
  kuzzle:
  # ...
  elasticsearch:
    image: yourImage
    environments:
      path.repo:
        - /tmp/snapshots
    # ...
```
or using regular YAML array of string:
```yaml
services:
  kuzzle:
  # ...
  elasticsearch:
    image: yourImage
    environments:
      - "path.repo=[/tmp/snapshots]"
    # ...
```

alexandrebouthinon: Double quote in the second snippet are required for YAML parsing

src/commands/es/snapshots/dump.ts
```diff
@@ -0,0 +1,46 @@
+import { flags } from '@oclif/command'
+import { Client } from '@elastic/elasticsearch'
+
+import { Kommand } from '../../../common'
+
+export default class EsSnapshotsDump extends Kommand {
```
alexandrebouthinon, line 6: I don't think we should use the term "dump" in this module since they are not dumps but snapshots which is quite different
```suggestion
export default class EsSnapshots extends Kommand {
```

alexandrebouthinon (changes requested):

src/commands/es/snapshots/dump.ts
Aschen: So the command should be called `create` to follow ES terminology: https://www.elastic.co/guide/en/elasticsearch/reference/current/snapshot-restore.html

src/commands/es/snapshots/restore.ts
rolljee: The restore operation must be performed on a functioning cluster. However, an existing index can be only restored if it’s closed and has the same number of shards as the index in the snapshot. The restore operation automatically opens restored indices if they were closed and creates new indices if they didn’t exist in the cluster.


Documentation is clear, we don't need to do that

features/docker/elasticsearch/elasticsearch.yml
rolljee: I'll try that, get back to you if it works or not 👍

src/commands/es/snapshots/dump.ts
rolljee: I think this is a little confusing to just call it `EsSnapshots`

Since in this "snapshots" module we can create 2 types of thing

1. Repository (the holder of snapshots)
1. Snapshot (data)

I prefer option `EsSnapshotsCreate` since we have an `EsSnapshotsCreateRepository`

@alexandrebouthinon wdyt ?

rolljee: And also, shall we rename the Kommand -> es:snapshots:create ?

ping @Aschen @alexandrebouthinon

features/docker/elasticsearch/elasticsearch.yml
rolljee: @alexandrebouthinon both of the options you proposed weren't successful. Only option I see is to keep it that way :x

alexandrebouthinon: Ok I thought it was a parsing and type inference problem. That's good for me then 🙂

src/commands/es/snapshots/dump.ts
alexandrebouthinon: Yes I agree with you for the naming, we might as well stick to the Elasticsearch documentation 👍

alexandrebouthinon (approved):

Aschen (approved):
Last nitpicking I'm sorry :sweat_smile: 
The commands should be under the name `snapshot` and not `snapshots` since that's the name of a "module"

PR #78, opened 2020-11-02
Fix app:scaffold and improve api action hook help

Aschen:
## What does this PR do?

Following @Nerwin reviews about the Getting Started, I've made few changes.

For the app:scaffold command:
 - add a docker-compose.yml to build and run the app
 - use docker-compose to install dependencies
 - some UX improvement

For the API action hook when no command is found:
 - when users tried to display the help of an inferred command (e.g. `kourou auth:getCurrentUser --help) they get the `command not found error`
 - I added a monkey patch to return the help of the hook instead
 
### Other change

 - automatically convert string into boolean for WebSocket protocol
 - `sdk:execute` command `--code` flag is now an argument


rolljee (approved):

bin/run
```diff
@@ -1,5 +1,34 @@
 #!/usr/bin/env node
 
+const { Config } = require('@oclif/config')
```
xbill82, line 3: Oh... Isn't this file automatically generated? Are you sure you can monkey-patch it without getting it overwritten on the next iteration?

src/commands/sdk/query.ts
```diff
@@ -4,7 +4,34 @@ import _ from 'lodash'
...
+
+  Kourou will try to infer the first arguments to one the following pattern:
+    - <command> <index>
+    - <command> <body>
+    - <command> <index> <collection>
+    - <command> <index> <collection> <id>
+    - <command> <index> <collection> <body>
+    - <command> <index> <collection> <id> <body>
+
+  If a flag is given (-i, -c, --body or --id), then the flag value has prior to
```
xbill82, line 24:
```suggestion
  If a flag is given (-i, -c, --body or --id), then the flag value has priority over
```

templates/app-scaffold/ts/package.json
```diff
@@ -4,7 +4,10 @@
   "description": "",
   "author": "",
   "scripts": {
-    "dev": "nodemon --ext 'js,json,ts' --inspect=0.0.0.0:9229 --exec node -r ts-node/register app.ts",
+    "npm:docker": "docker-compose run kuzzle npm",
```
xbill82, line 7: Just `npm`?

xbill82: ah, I guess the purpose of this is to be used by other scripts...

bin/run
Aschen: When you type `kourou auth:getCurrentUser`, the custom `command_not_found` hook is executed and we can use the `sdk:query` command with the correct controller and action.

When you type `kourou auth:getCurrentUser --help`, the `@ocliff/plugin-help` take precedence and use the `Config` class from the `@ocliff/config` package and since the command does not exists it displays the `command not found` error.

They cannot overwrite this monkey patch but if they release a new version of `@ocliff/config` then it may cause issues but they don't seems to update this [part of the code](https://github.com/oclif/config/blob/master/src/config.ts#L384) often.

templates/app-scaffold/ts/package.json
Aschen: Command suffixed with `:docker` are ran inside the container. For example you may want to bump the version, update dependencies, etc

bin/run
xbill82: 👍

xbill82 (approved):

bin/run
```diff
@@ -1,33 +1,30 @@
 #!/usr/bin/env node
 
-const { Config } = require('@oclif/config')
-
-const SdkQuery = require('../lib/commands/sdk/query')
+const { default: Help } = require('@oclif/plugin-help')
 
+const Help_showHelp = Help.prototype.showHelp;
```
xbill82, line 5: I don't want to firmly oppose to this monkey-patching. After all, it works and will do (as long as we don't update the Help plugin _and_ its API changes). But I think that the proper way to implement the solution to our need is to add our own help plugin.
https://oclif.io/docs/plugins

Aschen: I agree, we could also fork the existing one and adapt it for our needs.   

The first monkey patching was quite dangerous because any API change would break it. This one call the original method, catch the error, enhance it and then throw it.  
The only difference between the original method and the patched one is the error message.

I choose monkey patching because this time because it was safe and quick

PR #87, opened 2020-11-18
Make sure that all dependencies for a scaffolded application are installed

scottinet:
# Description

The `app:scaffold` command only installs the `kuzzle` dependency, namely, effectively ignoring the dependencies listed in the `package.json` file.

This prevents the `dev:docker` command to run correctly since the `ts-node` module is not installed when the application is scaffolded.

This PR fixes that.

# Other changes

More accurate README.md file for the scaffolded application

### DIFF
--- a/src/commands/app/scaffold.ts
+++ b/src/commands/app/scaffold.ts
@@ -35,7 +35,7 @@ export default class AppScaffold extends Kommand {
 
     const tasks = new Listr([
       {
-        title: `Creating ${chalk.gray(this.args.name + path.sep)} directory`,
+        title: `Creating directory: ${chalk.gray(this.args.name + path.sep)}`,
         task: () => execute('mkdir', this.args.name)
       },
       {
@@ -43,15 +43,15 @@ export default class AppScaffold extends Kommand {
         task: () => this.renderTemplates(templatePath, templatePath)
       },
       {
-        title: 'Installing latest Kuzzle version via NPM and Docker (this can take some time)',
+        title: 'Installing dependencies (this can take some time)',
         task: () =>
-          execute('npm', 'run', 'npm:docker', 'install', 'kuzzle', { cwd: this.args.name })
+          execute('npm', 'run', 'install:docker', { cwd: this.args.name })
       },
     ])
 
     await tasks.run()
 
-    this.logOk(`Scaffolding complete! Use "${chalk.grey('npm run dev:docker')}" to run your application`)
+    this.logOk(`Scaffolding complete! Use "${chalk.grey('cd ' + this.args.name + ' && npm run dev:docker')}" to run your application`)
   }
 
   async renderTemplates(directory: string, templatePath: string) {
--- a/templates/app-scaffold/ts/README.md
+++ b/templates/app-scaffold/ts/README.md
@@ -6,14 +6,18 @@ _An application running with [Kuzzle](https://github.com/kuzzleio/kuzzle)_
 
 ### Installation
 
-Requirement: 
- - Docker
- - Docker-Compose
+The command `kourou app:scaffold` takes care of installing dependencies for you, so you shouldn't need to worry about it.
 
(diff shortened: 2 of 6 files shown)


templates/app-scaffold/ts/README.md
```diff
@@ -6,14 +6,18 @@ _An application running with [Kuzzle](https://github.com/kuzzleio/kuzzle)_
...
+However, if for some reason you need to reinstall dependencies (or add new ones), then you can easily do so by running the following command:
+
+`npm run install:docker`
+
+This command installs the dependencies listed in the `package.json` file of this project, inside the docker image. Since some dependencies are written in C or in C++ and compiled, installing in the target docker image ensures that there won't be incompatibility problems if your current system is different from the one used by docker to execute Kuzzle.
 
 ### Run
 
-You can use the `npm run dev:docker` to run the application with it's dependencies.  
+You can use the following command  to run your application in development mode:
```
Leodau, line 19:
```suggestion
You can use the following command to run your application in development mode:
```

Leodau:
`[✔] Kuzzle 2.7.0 is ready`
`info: 2020-11-18T14:39:01+00:00 [LOG:INFO] 'Application started'`
:+1:

Leodau (approved):

src/commands/app/scaffold.ts
```diff
@@ -35,23 +35,23 @@ export default class AppScaffold extends Kommand {
...
       {
         title: 'Creating and rendering application files',
         task: () => this.renderTemplates(templatePath, templatePath)
       },
       {
-        title: 'Installing latest Kuzzle version via NPM and Docker (this can take some time)',
+        title: 'Installing dependencies (this can take some time)',
         task: () =>
-          execute('npm', 'run', 'npm:docker', 'install', 'kuzzle', { cwd: this.args.name })
+          execute('npm', 'run', 'install:docker', { cwd: this.args.name })
```
Aschen, line 48: But then here `kuzzle` is not installed, the dependency is not listed in the scaffolded `package.json` since we want to always install the latest version

Aschen (changes requested):

src/commands/app/scaffold.ts
Aschen: We should first run `npm run install:docker` to install dependencies and then run `npm run npm:docker install kuzzle` to install Kuzzle latest version

scottinet: I want to check something first... I think that we can simply specify a major version for a NPM dependency. If we can do that, then we can let NPM do the job for us by simpling configuring the package.json file differently.

scottinet: @Aschen > the last commit should make sure that NPM installs the latest kuzzle v2 version when scaffolding a new application. Can you check it out?

templates/app-scaffold/ts/README.md
```diff
@@ -6,14 +6,18 @@ _An application running with [Kuzzle](https://github.com/kuzzleio/kuzzle)_
...
 
-Requirement: 
- - Docker
- - Docker-Compose
+The command `kourou app:scaffold` takes care of installing dependencies for you, so you shouldn't need to worry about it.
 
-First you need to install dependencies. You can use the `npm run install:docker` to install dependencies from inside the container.
+However, if for some reason you need to reinstall dependencies (or add new ones), then you can easily do so by running the following command:
+
+`npm run install:docker`
```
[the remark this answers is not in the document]
Aschen: I wanted to differentiate commands that are going to run with docker and those who are going to run on the host.  

But maybe the naming is wrong

Aschen (approved):

templates/app-scaffold/ts/README.md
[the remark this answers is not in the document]
scottinet: I agree that this name is misleading. This should change in a dedicated PR though.

PR #93, opened 2021-01-19
Update app template

Aschen:
## What does this PR do?

Update the application template:
 - bump dependencies
 - add github actions
 - add functional test framework


scottinet (approved):

templates/app-scaffold/ts/.github/workflows/pull_request.workflow.yml
```diff
@@ -0,0 +1,47 @@
+name: Pull request checks
```
xbill82, line 1: is it a good idea to scaffold the github actions? Just wondering.

xbill82 (approved):

templates/app-scaffold/ts/.github/workflows/pull_request.workflow.yml
Aschen: IMHO Github actions are part of the standard Kuzzle workflow which includes a linter and functional tests (like Vue.js/Webpack does if I remember well)

If they don't want to use it they can just delete the files but IMHO users shouldn't

alexandrebouthinon: I partly agree with Luca. Maybe we should however remove the deployment part which can be very different depending on the use case

templates/app-scaffold/ts/.github/workflows/push_production.workflow.yml
```diff
@@ -0,0 +1,95 @@
...
+            registry: registry.gitlab.com
+            username: ${{ secrets.GITLAB_DOCKER_USERNAME }}
+            password: ${{ secrets.GITLAB_DOCKER_TOKEN }}
+        - name: Build and push image
+          uses: docker/build-push-action@v2
+          with:
+            context: ./backend
+            file: ./backend/Dockerfile
+            push: true
+            tags: registry.gitlab.com/kuzzle/<CHANGE-ME>:latest
```
Leodau, line 72: Is there a place where we give instructions to come here and update these **<CHANGE-ME>** ?

Aschen: No, but as we agreed, the deployment part is now removed

PR #114, opened 2021-10-29
Adding the Open API specification

clementbolin:
⚠️ Depends on [#2165](https://github.com/kuzzleio/kuzzle/issues/2165)

## What does this PR do ?

This PR adds an initial implementation of swagger-ui and describes the document controller's counting action.

```npm run doc:swagger```

### Other changes

Add docker-compose file to create a container with swagger image
Add package.json script to run the swagger ui.
Add README.md file for my ressources

PS: I will delete the launch script for swagger, as soon as I finish the documentation and add a command to kourou

### DIFF
--- a/doc/swagger/docker-compose.yaml
+++ b/doc/swagger/docker-compose.yaml
@@ -1,0 +1,11 @@
+version: '3.3'
+services:
+  swagger-ui:
+    image: swaggerapi/swagger-ui
+    container_name: swagger-ui
+    ports:
+      - "8001:8080"
+    volumes:
+      - ./swagger:/usr/share/nginx/html/swagger
+    environment:
+      API_URL: swagger/api.yaml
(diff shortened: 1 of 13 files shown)


Aschen (changes requested):
Even if Docker Compose is handy, it's more difficult to package since you need to ship a file with Kourou in the NPM package.  

It would be more convenient to just run a docker container with the correct configuration, especially since you only have one container to run.

Also, why the Open API definition files are part of this PR? The Open API definition should be loaded from Kuzzle HTTP `server:openApi` route.

Finally, rather than adding a script in the `package.json` file, you should add a new command to Kourou

Aschen:
The develop branch was late compare to master because we forgot to merge back master on develop after the last release. I fixed that and updated your branch so only your changes are seen on this PR

xbill82 (changes requested):
Neither I really understand this PR. Why is it bearing all the routes definitions? These should be in the Kuzzle repo, or am I missing something?

Aschen:
@clementbolin Any updates on this?

PR #118, opened 2021-11-26
Dump collection to CSV

xbill82:
Enables to dump a collection to CSV by choosing a set of fields, like the following example

```
kourou collection:export nyc-open-data:yellow-taxi --format csv --fields city,district,nested.field
```

You can create a collection with some documents and run the above command ☝️ 

### Other changes

A bit of refactoring was necessary to keep things clean. Dump is now done through abstract class inheritance and sub-classes are specialized on different formats.

### Boyscout

ESLint rules are now less upsetting for functional test files.


src/common.ts
```diff
@@ -173,3 +173,15 @@ export abstract class Kommand extends Command {
...
+/**
+ * An iteration-order-safe version of lodash.values
+ *
+ * @param object The object containing the values
+ * @param fields The field names to pick in the right order
+ * @returns The values in the same order as the fields
+ * @see https://lodash.com/docs/4.17.15#values
+ */
+ export function pickValues(object: any, fields: string[]): any[] {
+  return fields.map(f => object[f])
```
Aschen: Are you handling this kind of path `path.to.property` somewhere else?

src/support/dump-collection.ts
```diff
@@ -141,98 +141,63 @@ class KuzzleDumper extends JSONLDumper {
...
-export async function dumpCollectionData(sdk: any, index: string, collection: string, batchSize: number, destPath: string, query: any = {}, format = 'jsonl') {
+class CSVDumper extends AbstractDumper {
+  constructor(
+    sdk: any,
+    index: string,
+    collection: string,
+    batchSize: number,
+    destPath: string,
+    query: any = {},
+    protected readonly fields: string[]
```
Aschen: I don't know this syntax, what does it does?

src/support/dump-collection.ts
```diff
@@ -141,98 +141,63 @@ class KuzzleDumper extends JSONLDumper {
...
+  ) {
+    super(sdk, index, collection, batchSize, destPath, query)
+  }
+
+  protected get fileExtension(): string {
+    return 'csv'
+  }
+  setup(): Promise<void> {
+    return new Promise(resolve => resolve())
+  }
```
Aschen: This could simply be an empty `async` function

src/support/dump-collection.ts
```diff
@@ -141,98 +141,63 @@ class KuzzleDumper extends JSONLDumper {
...
+    }
+    this.fields.unshift('_id')
+    return this.writeLine(this.fields.join(','))
+  }
+  writeLine(line: any): Promise<void> {
+    if (!this.writeStream) {
+      throw new Error('Cannot write data: WriteStream is not initialized.')
+    }
+    this.writeStream.write(line)
+    return new Promise(resolve => resolve())
```
Aschen, line 179: This method does not need a promise, since you are using streams, the asynchronous operations are done automatically

src/support/dump-collection.ts
```diff
@@ -141,98 +141,63 @@ class KuzzleDumper extends JSONLDumper {
...
+  }
+  onResult(document: { _id: string; _source: any }): Promise<void> {
+    const values = [document._id, ...pickValues(document._source, this.fields)]
+    return this.writeLine(values)
+  }
+  tearDown(): Promise<void> {
```
Aschen: Same comment about `async` method. Also maybe you could add those empty methods in `AbstractDumper` class and then child class can implements a new version or not

README.md
```diff
@@ -379,31 +379,36 @@ ARGUMENTS
...
-  --batch-size=batch-size  [default: 2000] Maximum batch size (see limits.documentsFetchCount config)
-  --editor                 Open an editor (EDITOR env variable) to edit the query before sending
+  --api-key=api-key          Kuzzle user api-key
+  --as=as                    Impersonate a user
+  --batch-size=batch-size    [default: 2000] Maximum batch size (see limits.documentsFetchCount config)
+  --editor                   Open an editor (EDITOR env variable) to edit the query before sending
 
-  --format=format          [default: JSONL] "jsonl or kuzzle - kuzzle will export in Kuzzle format usable for internal
-                           fixtures and jsonl allows to import that data back with kourou
+  --fields=fields            [CSV format only] The fields to be included in the CSV export in dot-path format (e.g.
```
Aschen, line 387: Those fields should be separated with a coma? If yes, you should update the example to be more clear

src/support/dump-collection.ts
```diff
@@ -3,86 +3,202 @@ import path from 'path'
...
-  const collectionDir = path.join(destPath, collection)
-  const filename = path.join(collectionDir, (format.toLowerCase() === 'jsonl' ? 'documents.jsonl' : 'documents.json'))
-  const writeStream = fs.createWriteStream(filename)
-  const waitWrite = new Promise(resolve => writeStream.on('finish', resolve))
-  const ndjsonStream = ndjson.stringify()
-  const options = {
-    scroll: '1m',
-    size: batchSize
+  public setup(): Promise<void> {
+    return new Promise(resolve => resolve())
```
Aschen, line 32: You could just add the `async` keyword instead of returning a promise like this

src/support/dump-collection.ts
```diff
@@ -3,86 +3,202 @@ import path from 'path'
...
+  protected ndjsonStream = ndjson.stringify()
+
+  setup(): Promise<void> {
+    this.ndjsonStream.on('data', (line: string) => {
+      if (!this.writeStream) {
+        throw new Error('Cannot write data: WriteStream is not initialized.')
+      }
+      this.writeStream.write(line)
+    })
+    return new Promise(resolve => resolve())
```
Aschen, line 101: You could just mark the method as `async`

src/support/dump-collection.ts
```diff
@@ -3,86 +3,202 @@ import path from 'path'
...
+    return 'json'
+  }
+  onResult(document: {_id: string, _source: any}): Promise<void> {
+    this.rawDocuments[this.index][this.collection].push({
+      index: {
+        _id: document._id
+      }
+    })
+    this.rawDocuments[this.index][this.collection].push(document._source)
+    return new Promise(resolve => resolve())
```
Aschen, line 140: You could just mark the method as `async`

Aschen (changes requested):

src/common.ts
xbill82: Yes, it is used by the `CSVDumper` in `collection:export`. But, since this function is very generic, I wanted to make it live in some sort of `utilities.js` file. Since it doesn't exist, I've put it in `common.js`.

src/support/dump-collection.ts
xbill82: Ah! Thank you! I was struggling with these signatures! 😅

README.md
xbill82: ah yes you are right

src/support/dump-collection.ts
Aschen: so good

src/support/dump-collection.ts
Aschen: Yes I agree with you then :+1:

src/support/dump-collection.ts
```diff
@@ -3,86 +3,234 @@ import path from 'path'
...
+    protected readonly sdk: any,
+    protected readonly index: string,
+    protected readonly collection: string,
+    protected readonly batchSize: number,
+    protected readonly destPath: string,
+    protected readonly query: any = {},
+  ) {
+    this.collectionDir = path.join(this.destPath, this.collection)
+    this.options = {
+      scroll: '1m',
```
Aschen, line 26: 1 minute is very long for a scroll page TTL. It should be the time to write all the documents of the page (2000 by default, 10 000 maximum with Kuzzle/ES) in a file.

src/support/dump-collection.ts
```diff
@@ -3,86 +3,234 @@ import path from 'path'
...
-    scroll: '1m',
-    size: batchSize
+  /**
+   * One-shot call before the dump. Can be used to
+   * perform setup operations before dumping.
+   *
+   * @returns void
+   */
+  public async setup() {
+    return
```
Aschen, line 38: You don't even need this `return`

src/support/dump-collection.ts
```diff
@@ -3,86 +3,234 @@ import path from 'path'
...
-  const waitWrite = new Promise(resolve => writeStream.on('finish', resolve))
-  const ndjsonStream = ndjson.stringify()
-  const options = {
-    scroll: '1m',
-    size: batchSize
+  /**
+   * One-shot call before the dump. Can be used to
+   * perform setup operations before dumping.
+   *
+   * @returns void
```
Aschen, line 35: Nitpicking: there is little interest to document the void return since the TS type will already indicate this

src/support/dump-collection.ts
```diff
@@ -3,86 +3,234 @@ import path from 'path'
...
-        })
-        rawDocuments[index][collection].push(hit._source)
+  async setup() {
+    this.ndjsonStream.on('data', (line: string) => {
+      if (!this.writeStream) {
+        throw new Error('Cannot write data: WriteStream is not initialized.')
       }
+      this.writeStream.write(line)
+    })
+    return
```
Aschen, line 133: useless return

src/support/dump-collection.ts
```diff
@@ -3,86 +3,234 @@ import path from 'path'
...
-  ndjsonStream.end()
-  writeStream.end()
+  protected get fileExtension(): string {
+    return 'csv'
+  }
+  setup(): Promise<void> {
+    if (this.fields.includes('_id')) {
+      this.fields.splice(this.fields.indexOf('_id'), 1)
+    }
+    return new Promise(resolve => resolve())
```
Aschen, line 200: useless promise, you should use`async` instead

src/support/dump-collection.ts
```diff
@@ -3,86 +3,234 @@ import path from 'path'
...
+  writeHeader(): Promise<void> {
+    return this.writeLine(['_id', ...this.fields].join(this.separator))
+  }
+  async writeLine(line: any) {
+    if (!this.writeStream) {
+      throw new Error('Cannot write data: WriteStream is not initialized.')
+    }
+    this.writeStream.write(line)
+    this.writeStream.write('\n')
+    return
```
Aschen, line 211: Useless `return`

Aschen (changes requested):

src/commands/collection/export.ts
```diff
@@ -51,6 +62,14 @@ export default class CollectionExport extends Kommand {
       ? path.join(this.flags.path, this.args.index)
       : this.args.index
 
+    const fields = this.flags.fields
+      ? this.flags.fields.split(',')
+      : []
+
+    if (this.flags.format === 'csv' && fields.length === 0) {
+      this.logInfo('It looks like you are exporting to CSV but you did not select any field. Only document IDs will be exported.')
```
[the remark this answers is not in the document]
xbill82: I've thought about it but, what if the user just wants to export the IDs?

src/common.ts
```diff
@@ -173,3 +175,30 @@ export abstract class Kommand extends Command {
...
+/**
+ * Formats the value for correct CSV output, avoiding to return
+ * values that would badly serialize in CSV.
+ *
+ * @param value The value to format
+ * @returns The value or a string telling the value is not scalar
+ */
+export function formatValueForCSV(value: any) {
+  if (isObject(value)) {
+    return '[NOT_SCALAR]'
```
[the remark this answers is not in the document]
xbill82: a stringified Array gives `[e1,e2,e3]`, a stringified Object gives `{a1:v1, a2:v2}` and both conflict with the CSV convention. Given that the users that will need this are fairly not power users, I don't feel comfortable with the idea of giving them output that will mess up their Excel import.

src/support/dump-collection.ts
xbill82: This is code I just re-used from your implementation. I'll change it to 2 seconds.

src/support/dump-collection.ts
xbill82: cool!

xbill82: 👎 https://github.com/kuzzleio/kourou/runs/4405960795?check_suite_focus=true#step:5:24

Aschen: You can ignore the warning then `// eslint-disable-next-line @typescript-eslint/no-empty-function`, that's what we are doing in other cases like this (e.g. `init: () => {}`)

src/commands/collection/export.ts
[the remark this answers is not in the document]
Aschen: Good question, honestly I think it's an edge case and most of the time users may want to export the content. But maybe if they only put `--fields=_id` then we can only dump ids?

src/common.ts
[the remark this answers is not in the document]
Aschen: I agree :+1: (CSV :facepalm: )

src/commands/collection/export.ts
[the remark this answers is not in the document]
xbill82: ok, let's do it like this

src/support/dump-collection.ts
xbill82: yes yes, I just wanted you to read me complaining about linting rules 😂

src/commands/collection/export.ts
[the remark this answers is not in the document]
xbill82: Too bad the `flattenObject` function is not exported by Koncorde...

Aschen (approved):

features/CSV.feature
```diff
@@ -28,3 +28,26 @@ Feature: CSV Export
...
+      | "the-hive-th"     | { "city": "changmai", "district": 7, "unwanted": "field" } |
+      | "luca"            | { "city": "changmai", "district": [7, 8, 9] }              |
+      | "chuon-chuon-kim" | { "city": "hcmc", "district": 1, "nested": {"field": 1} }  |
+    And I refresh the collection
+    # collection:export
+    When I run the command "collection:export" with:
+      | arg  | nyc-open-data |     |
+      | arg  | yellow-taxi   |     |
+      | flag | --format      | csv |
+    Then I get the file in "nyc-open-data/yellow-taxi/documents.csv" containing
```
Aschen, line 46: This test does not prove that nested fields are correctly dumped, can you add an example of this?

Aschen (changes requested):

features/CSV.feature
xbill82: This is tested in the [previous feature](https://github.com/kuzzleio/kourou/pull/118/files/83de276ebb49c1c7eed81451a37317f6030edc8c..6f339cd90e25cc55470236e8ba510ad94dbba2bd#diff-81fb5b99c2065787b6b50d999406c43a52ca73efe2c7ee8cbf1e966d0b200f8dR27)

src/commands/collection/export.ts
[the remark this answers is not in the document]
Aschen: Even if it was, it's a huge dependency to include just for this :sweat_smile:

features/CSV.feature
Aschen: oh yes, sorry :+1:

Aschen (approved):

PR #130, opened 2022-02-16
Add commands to interact with the PaaS

Aschen:
## What does this PR do?

Add 3 commands:
 - `paas:init`: add the special entry in `package.json` to identify paas related project and their namespace
 - `paas:login`: login to the specified namespace and save credentials in Kourou folder
 - `paas:publish`: publish a new version of the application

I didn't had time to add tests and it will be a little bit tricky since we need to mock the PaaS.
I lack of time to add them now but they will be added for the first public release


rolljee (changes requested):

src/commands/paas/publish.ts
```diff
@@ -30,38 +26,50 @@ class PaasPublish extends Kommand {
...
+    await this.initPaasClient({ apiKey });
+
+    const user = await this.paas.auth.getCurrentUser();
+    this.logInfo(`Logged as "${user._id}" for project "${this.args.project}"`);
+
+    const [image, tag] = this.args.image.split(':');
+    this.logInfo(`Deploy application with image "${image}:${tag}"`)
+
+    await this.paas.query({
+      controller: 'github',
```
rolljee, line 40: This will soon be updated to `config`

```suggestion
      controller: 'config',
```

rolljee (changes requested):

rolljee (approved):

PR #135, opened 2022-03-08
Fix issue with scroll duration & improve flag

rolljee:
## What does this PR do?

* Fix issue where scrollId parameters was not taken into account
* Improve migrate command support with dry-run option, regex pattern and specific scroll duration for large collection support

  - Step 1 : Start two elasticsearch
  - Step 2 : Run this command `./bin/run es:migrate --src http://localhost:9201 --dest http://localhost:9200 --dry-run --pattern i-want-to-keep-this`
  - Step 3 :  See command output
  ...

### DIFF
--- a/src/commands/es/migrate.ts
+++ b/src/commands/es/migrate.ts
@@ -7,129 +7,151 @@ import emoji from 'node-emoji'
 import { Kommand } from '../../common'
 
 export default class EsMigrate extends Kommand {
-    static initSdk = false
+  static initSdk = false
 
-    static description = 'Migrate all the index from one Elasticsearch server to another'
+  static description = 'Migrate all the index from one Elasticsearch server to another'
 
-    static flags = {
-        help: flags.help(),
-        src: flags.string({
-            description: 'Source Elasticsearch server URL',
-            env: 'KUZZLE_ES_SRC',
-            required: true
-        }),
-        dest: flags.string({
-            description: 'Destination Elasticsearch server URL',
-            env: 'KUZZLE_ES_DEST',
-            required: true
-        }),
-        reset: flags.boolean({
-            description: 'Reset destination Elasticsearch server',
-            default: false,
-        }),
-        'batch-size': flags.integer({
-            description: 'How many documents to move in batch per operation',
-            default: 1000,
-        }),
-        'no-interactive': flags.boolean({
-            description: 'Skip confirmation interactive prompts (perfect for scripting)',
-            default: false,
-        }),
+  static flags = {
+    help: flags.help(),
+    src: flags.string({
+      description: 'Source Elasticsearch server URL',
+      env: 'KUZZLE_ES_SRC',
+      required: true
+    }),
+    dest: flags.string({
+      description: 'Destination Elasticsearch server URL',
+      env: 'KUZZLE_ES_DEST',
+      required: true
+    }),
+    reset: flags.boolean({
+      description: 'Reset destination Elasticsearch server',
+      default: false,
+    }),
+    'batch-size': flags.integer({
+      description: 'How many documents to move in batch per operation',
+      default: 1000,
+    }),
+    'no-interactive': flags.boolean({
+      description: 'Skip confirmation interactive prompts (perfect for scripting)',
+      default: false,
+    }),
+    'dry-run': flags.boolean({
+      description: 'Print witch collections will be migrated',
+      default: false,
+    }),
+    'pattern': flags.string({
+      description: 'Pattern to match indices to migrate'
+    }),
+    'scroll': flags.string({
+      description: 'Scroll duration for Elasticsearch scrolling',
+      default: '30s'
+    }),
+  }
+
+  static examples = [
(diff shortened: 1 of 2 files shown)


src/commands/es/migrate.ts
```diff
@@ -7,129 +7,151 @@ import emoji from 'node-emoji'
...
+
+  static examples = [
+    'kourou es:migrate --src http://elasticsearch:9200 --dest http://otherElasticsearch:9200 --reset --batch-size 2000',
+    'kourou es:migrate --src http://elasticsearch:9200 --dest http://otherElasticsearch:9200 --reset --batch-size 2000 --no-interactive'
+  ]
+
+  private async migrateMappings(index: string, src: Client, dest: Client) {
+    const { body } = await src.indices.getMapping({ index })
+    const mappings = body[index]
+    await dest.indices.create({ index, body: { mappings: mappings.mappings, settings: { number_of_shards: 1, number_of_replicas: 0 } } })
```
Aschen, line 59: Why hardcoded `number_of_shards` and replicas?

src/commands/es/migrate.ts
```diff
@@ -7,129 +7,151 @@ import emoji from 'node-emoji'
...
+  }
+
+  private async migrateData(index: string, batch_size: number, src: Client, dest: Client) {
+    const queue = []
+
+    const { body } = await src.search({
+      index,
+      scroll: this.flags.s,
```
Aschen, line 67: Use entire name
```suggestion
      scroll: this.flags.scroll,
```

src/commands/es/migrate.ts
```diff
@@ -7,129 +7,151 @@ import emoji from 'node-emoji'
...
 
-    static examples = [
-        'kourou es:migrate --src http://elasticsearch:9200 --dest http://otherElasticsearch:9200 --reset --batch-size 2000',
-        'kourou es:migrate --src http://elasticsearch:9200 --dest http://otherElasticsearch:9200 --reset --batch-size 2000 --no-interactive'
-    ]
+    const progressBar = cli.progress({ format: chalk.blue(' [*] Importing |{bar}| {percentage}% || {value}/{total} documents') })
+    progressBar.start(total, 0)
+
+    let count = 0
+    queue.push(body.hits)
```
Aschen, line 86: Instead using a one element queue, you could simply have a `let hits = null` variable where you store current page hits

Aschen (changes requested):

src/commands/es/migrate.ts
rolljee: Oh, my bad this was for a one shot migration, i'll make this a flag too

Aschen: I think it should rather be the former values. Otherwise if you want to make it configurable then I think a config file is more appropriate since you may have many indices

Aschen (approved):

src/commands/es/migrate.ts
```diff
@@ -7,129 +7,151 @@ import emoji from 'node-emoji'
...
+
+  static examples = [
+    'kourou es:migrate --src http://elasticsearch:9200 --dest http://otherElasticsearch:9200 --reset --batch-size 2000',
+    'kourou es:migrate --src http://elasticsearch:9200 --dest http://otherElasticsearch:9200 --reset --batch-size 2000 --no-interactive'
+  ]
+
+  private async migrateMappings(index: string, src: Client, dest: Client) {
+    const { body } = await src.indices.getMapping({ index })
+    const mappings = body[index]
+    await dest.indices.create({ index, body: { mappings: mappings.mappings } })
```
alexandrebouthinon, line 59:
```suggestion
    await dest.indices.create({ index, body: { mappings: mappings.mappings } })
```
I think you should migrate index settings too

Aschen: Beware of settings that cannot be changed

rolljee: There is two way to approach this

1. We only allow a list of settings for the user
2. We let the user decide witch setting he want and take care of consequences that might happen with it

For me the option 1 is better and so accept this as is, make a separate PR and propose a moment to discuss witch are the settings we will allow

wdyt @Aschen @alexandrebouthinon  ?

Aschen: I'm not fond of the solution where we have to maintain a list to be honest, it will be difficult to keep it up to date with ES, also specifics of the underlying storage engine should not be that much of our concern

alexandrebouthinon: I don't know IMO es:migrate aim to move index as it come from one Elasticsearch to another without updating it (mappings AND settings). You should take the source index settings and use them without modify/moderate them

Aschen: I agree but then the destination index should not exists beforehand

alexandrebouthinon: Maybe we could update the command source code to make it test all index from source Elasticsearch don't exist in the destination one. But I think we're out of the scope of this PR, let me create an associated issue 😉
https://github.com/kuzzleio/kourou/issues/142

PR #140, opened 2022-05-16
Add configurable scrollTTL for `collection:export`

xbill82:
Added `--scrollTTL` flag to `collection:export`

### DIFF
--- a/src/support/dump-collection.ts
+++ b/src/support/dump-collection.ts
@@ -66,10 +66,11 @@ abstract class AbstractDumper {
     protected readonly batchSize: number,
     protected readonly destPath: string,
     protected readonly query: any = {},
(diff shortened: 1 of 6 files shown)


src/support/dump-collection.ts
```diff
@@ -227,9 +228,11 @@ class CSVDumper extends AbstractDumper {
     destPath: string,
     query: any = {},
     protected fields: string[],
-    protected readonly separator = ','
+    protected readonly separator = ',',
+    protected readonly scrollTTL: string = '2s'
```
OlivierCavadenti, line 232: init here is not necessary since we put the same default value in the parent

Aschen (approved):

PR #143, opened 2022-05-31
Add a command to make a proxy that allows the chrome inspector to use the DebugController

Shiranuit:
depends on https://github.com/kuzzleio/kuzzle/pull/2358
## What does this PR do?

This PR adds the command `kourou app:debug-proxy` that creates a proxy that allows the Chrome Inspector to connects to Kuzzle and debug Kuzzle through the DebugController api actions.


rolljee (approved):
lgtm

PR #148, opened 2022-07-07
Allow user to select the applicationId he want to deploy

rolljee:
# Overview

This feature is needed for the new paas possibility of having multiple kuzzle in an env

### DIFF
--- a/src/commands/paas/deploy.ts
+++ b/src/commands/paas/deploy.ts
@@ -22,6 +22,7 @@ class PaasDeploy extends PaasKommand {
 
   static args = [
     { name: 'environment', description: 'Project environment name', required: true },
+    { name: 'applicationId', description: 'Application Identifier', required: true },
     { name: 'image', description: 'Image name and hash as myimage:mytag', required: true },
   ]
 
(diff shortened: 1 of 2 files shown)


Shiranuit (approved):

alexandrebouthinon (commented):
To avoid breaking previously established CI/CD we should use `kuzzle` as the default value for this new argument

Aschen (changes requested):
I agree with @alexandrebouthinon comment, default value should be `kuzzle`

rolljee:
@Aschen @alexandrebouthinon 

This is pointless, it will not work anyway even with the default value as kuzzle.

The new process now use `_id` witch is choosen by the user, kuzzle was the property `name`.

The code does not use the property name anymore.

We will need to go through each individual CI/CD to modify it anyway.

Aschen (commented):
I leave you this PR since I don't have enough knowledge to review it

rolljee:
@Aschen I'll need you to accept the PR so i can merge it :)

Aschen (approved):

PR #149, opened 2022-07-26
Add options to control login ttl during debugging

Shiranuit:
## What does this PR do?

Add options to control the login TTL and enable the automatic refresh of the token while debugging with the debug-proxy

### DIFF
--- a/src/commands/app/debug-proxy.ts
+++ b/src/commands/app/debug-proxy.ts
@@ -25,9 +25,17 @@ export default class DebugProxy extends Kommand {
       description: 'Port of the forwarding server',
       default: 9222
     }),
-    nonoAutoEnableDebugger: flags.boolean({
+    ttl: flags.string({
+      description: 'Kuzzle login TTL',
+      default: '1h',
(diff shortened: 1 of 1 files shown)


Aschen (approved):

src/commands/app/debug-proxy.ts
```diff
@@ -25,9 +25,17 @@ export default class DebugProxy extends Kommand {
...
       default: 9222
     }),
-    nonoAutoEnableDebugger: flags.boolean({
+    ttl: flags.string({
+      description: 'Kuzzle login TTL',
+      default: '1h',
+    }),
+    keepAuth: flags.boolean({
+      description: 'Keep the user authenticated',
+      default: false,
```
alexandrebouthinon, line 34: I can't see where you're using those two new flags in the command code

Aschen: They are used by Kourou in the main class

alexandrebouthinon (approved):

PR #157, opened 2022-09-12
Prevent cluster node eviction while debugging

Shiranuit:
## What does this PR do?

- Ask Kuzzle to prevent the node from being evicted while debugging when starting the debug-proxy
- Prevent the SDK from disconnecting itself because of the websocket `ping pong` while debugging because Kuzzle can be slow to respond


Shiranuit:
Eviction prevention is enabled when the debugger is enabled on a node automatically

PR #162, opened 2022-10-19
Add app:doctor command

### LINKED ISSUE #85: New command kourou app:doctor
In the same way Flutter provide the super convenient `flutter doctor` command to perform checks before development begin (Android SDK version, Android emulator version, Dart version...), we decided that we needed our own command.

Checks and information retrieved need to be discuss but IMO we need at least:
* Node.js:
    * Installed and configured (binary file in `$PATH` for example)
    * Version
* Kuzzle version check
* Elasticsearch:
    * Running
    * Version
    * Kernel settings (`vm.max_map_count=262144`)
* Redis:
    * Running
    * Version

Let's talk about it below 😄

OlivierCavadenti:
- Check some configurations and cases for a Kuzzle stack.
- List of checks:
   * admin exists or not
   * cluster is enabled
   * Redis and ElasticSearch running
   * Kuzzle, NodeJS and ElasticSearch versions. Checks if the version are supported by Kuzzle.
   * Show ElasticSearch cluster name, version, nodes and indices.
   * Checks installation of related Kuzzle system libraries (wip: only on Linux platform).
   * Check docker compose version
   * Show suggestions for each errors.

Example of output:
image

Closes #85


rolljee (approved):

src/commands/app/doctor.ts
```diff
@@ -0,0 +1,240 @@
...
+  };
+
+  static args = [];
+
+  static readStdin = true;
+
+  ELK_MAX_VERSION = "7.17";
+  NODEJS_MAX_VERSION = "14";
+
+  COMMANDS_PLATFORMS: { [key: string]: string } = {
```
Aschen, line 28: Since this is not modified, it should be defined as const without explicit type and then TS will automatically use keys present inside as type definition

https://github.com/kuzzleio/kuzzle-device-manager/blob/2-stable/features/fixtures/application/decoders/DummyTempDecoder.ts#L11

src/commands/app/doctor.ts
```diff
@@ -0,0 +1,240 @@
...
+    linux: "which",
+  };
+
+  LIBRARIES_PLATFORMS: { [key: string]: string[] } = {
```
Aschen, line 32: Since this is not modified, it should be defined as const without explicit type and then TS will automatically use keys present inside as type definition

https://github.com/kuzzleio/kuzzle-device-manager/blob/2-stable/features/fixtures/application/decoders/DummyTempDecoder.ts#L11

src/commands/app/doctor.ts
```diff
@@ -0,0 +1,240 @@
...
+  };
+
+  async runSafe() {
+    const suggestions = [];
+    const nodeVersion = await this.sdk.query({
+      controller: "debug",
+      action: "nodeVersion",
+    });
+
+    this.log(`----------------- Doctor begin his job ! -----------------`);
```
Aschen, line 55: Isn't it `DoKtor` ? ;p

src/commands/app/doctor.ts
```diff
@@ -0,0 +1,240 @@
...
+    this.log(`General checks`);
+
+    const adminExists = await this.sdk.server.adminExists({});
+    const anonymous = await this.sdk.security.getRole("anonymous");
```
Aschen, line 59: Those 3 request can be done in // (those 2 and the `nodeVersion` for example)

src/commands/app/doctor.ts
```diff
@@ -0,0 +1,240 @@
...
+
+    // Redis
+    if ("PONG" === (await this.sdk.ms.ping())) {
+      this.logOk("Redis Memory Storage is running");
+    } else {
+      this.logKo("=> Redis Memory Storage is not running");
+      suggestions.push("Check Redis instance dedicated to memory storage");
+    }
+
+    this.logOk("Kuzzle Version: " + config.version);
```
Aschen, line 127: You should use string interpolation instead (this remark apply everywhere)

src/commands/app/doctor.ts
```diff
@@ -0,0 +1,240 @@
...
+    this.logOk("Kuzzle NodeJS version: " + nodeVersion.result);
+    if (nodeVersion.result > this.NODEJS_MAX_VERSION) {
+      this.logKo(
+        "=> Kuzzle NodeJS version is not compatible with Kuzzle (max version is " +
+          this.NODEJS_MAX_VERSION +
+          ")"
+      );
+    }
+    this.logOk("NodeJS Building version: " + process.version);
+    if (nodeVersion.result != process.version) {
```
Aschen, line 137:
```suggestion
    if (nodeVersion.result !== process.version) {
```

src/commands/app/doctor.ts
```diff
@@ -0,0 +1,240 @@
...
+        "=> Kuzzle node version is different from the build node version"
+      );
+      suggestions.push(
+        `Match Kuzzle NodeJS version ${nodeVersion.result} with the build NodeJS version ${process.version}`
+      );
+    }
+
+    let elkVersion;
+    try {
+      const client = new Client({ node: this.flags.elasticsearch });
```
Aschen, line 148: ES version should rather be fetch from `server:info` because it's quite rare to expose ES directly apart when developing locally.

Or maybe this whole part should be optionnal

Aschen (changes requested):

src/commands/app/doctor.ts
OlivierCavadenti: I try that, but I have this compilation problem:

TS7053: Element implicitly has an 'any' type because expression of type 'string' can't be used to index type '{ readonly linux: "which"; }'.   No index signature with a parameter of type 'string' was found on type '{ readonly linux: "which"; }'.

Aschen: Thanks ocliff <3 Can you add a comment explaining why then?

src/commands/app/doctor.ts
OlivierCavadenti: I fetch es version from server:info and try to fetch other data from ES with the client. If not possible, I throw an error.

src/commands/app/doctor.ts
```diff
@@ -22,14 +22,15 @@ export default class AppDoctor extends Kommand {
 
   static readStdin = true;
 
-  ELK_MAX_VERSION = "7.17";
-  NODEJS_MAX_VERSION = "14";
+  ELK_MAX_VERSION = 7.17;
+  NODEJS_MAX_VERSION = 14;
```
Aschen, line 26: Officially we support Node.js 16

Aschen (approved):

PR #164, opened 2022-10-20
Add command es:aliases:cat

### LINKED ISSUE #108: New command es:aliases:cat
Since this [PR](https://github.com/kuzzleio/kuzzle/pull/2117), Kuzzle uses aliases instead of directly using indices when dealing with ES. Even though both are very similar for now, they may evolve independently in the future. That is why kourou should provide a command to list aliases similar to the one currently listing indices.
Once this new command has been implemented, Kuzzle could suggest it in the [doc](https://github.com/kuzzleio/kuzzle/pull/2117/files#diff-3ce338c203584cdab38f31fb6375a91655dfdd0b868664d14010911e6d47f609R60) section [Internal Representation](https://docs.kuzzle.io/core/2/guides/main-concepts/data-storage/#internal-representation)

OlivierCavadenti:
- Add command to list ES aliases and all associated indexes.

Close #108


rolljee:
This shell be tested with functional test

rolljee (changes requested):
Missing functional tests

Aschen (approved):

rolljee (approved):

PR #165, opened 2022-10-20
Various fixes and refacto for Kourou 1.0

### LINKED ISSUE #88: Kourou v1.0
## Kourou v1.0

Here is the list of needed changes before releasing the v1.0

 - [ ] `api-key:create` : The `description` flag should be an argument
 - [ ] `api-key:search` : The `filter` flag should be named `query`
 - [ ] `app:start-services` : Should be renamed in `services:start` and a flag should be added to specify Docker Compose project name
 - [ ] `config:diff` : Bug when displaying differences between object values (with the `--values` flag)
 - [ ] `document:search` : The `aggregations` flag should be added (maybe just name it `aggs`)
 - [ ] `*:export` : Export commands should all be in the `export:` namespace (e.g. `index:export` => `export:index`)
 - [ ] `*:import` : Import commands should be deleted and only the global `import` command should be kept
 - [ ] `*` : All commands using Kuzzle API should accept the `--arg` flag to pass arbitrary arguments in the request
 - [ ] `instance:*` : Instance commands should be removed since they are obsolete with kaaf
 - [ ] `redis:list-keys` : There is a bug in the command when no matching key is found
 - [ ] `kourou sdk:execute` : Should be renamed to `kourou execute`
 - [ ] `export:user` : A `--mappings` flag should be added to export mappings (and the former mappings related commands should be removed)
 - [ ] `sdk:query` : Should be renamed to `api` and the automatic inference of commands as API action should be removed

OlivierCavadenti:
Fixes:
- config:diff : Bug when displaying differences between object values (with the --values flag)
- redis:list-keys : There is a bug in the command when no matching key is found

Renaming:
- api-key:create : The description flag should be an argument
- api-key:search : The filter flag should be named query
- kourou sdk:execute : Should be renamed to kourou execute
- sdk:query : Should be renamed to api and the automatic inference of commands as API action should be removed
- app:start-services : Should be renamed in services:start and a flag should be added to specify Docker Compose project name

Flag adds:
- document:search : The aggregations flag should be added (maybe just name it aggs)
- export:user : A --mappings flag should be added to export mappings (and the former mappings related commands should be removed)

Removed:
- *:export : Export commands should all be in the export: namespace (e.g. index:export => export:index)
- *:import : Import commands should be deleted and only the global import command should be kept
- instance:* : Instance commands should be removed since they are obsolete with kaaf

Closes #88


features/Collection.feature
```diff
@@ -12,30 +12,14 @@ Feature: Collection Commands
...
+    When I run the command "export:collection" with:
       | arg  | nyc-open-data |                            |
       | arg  | yellow-taxi   |                            |
       | flag | --query       | { term: { city: "hcmc" } } |
       | flag | --scrollTTL   | 3s                         |
     Then I successfully call the route "collection":"delete" with args:
       | index      | "nyc-open-data" |
       | collection | "yellow-taxi"   |
-    # collection:import
-    And I run the command "collection:import" with args:
```
Aschen, line 24: Can you keep this test and use the `import` command instead to validate importing a collection is still working?

features/Index.feature
```diff
@@ -19,26 +19,3 @@ Feature: Index Commands
       | "nyc-open-data" |
     And I successfully call the route "index":"delete" with args:
       | index | "nyc-open-data" |
-    # index:import
-    And I run the command "index:import" with args:
```
Aschen, line 23: Same command, keep the test with `import` command

features/Profile.feature
```diff
@@ -1,22 +1,13 @@
...
     Given I create a profile "teacher" with the following policies:
       | default | [{ "index": "nyc-open-data" }] |
     And I create a profile "student" with the following policies:
       | admin | [{ "index": "mtp-open-data" }] |
-    When I run the command "profile:export" with:
+    When I run the command "export:profile" with:
       | flag | --path | ./dump |
-    And I delete the profile "teacher"
-    And I delete the profile "student"
-    When I run the command "profile:import" with:
```
Aschen, line 15: Same comment about `import` command

features/Role.feature
```diff
@@ -10,13 +10,5 @@ Feature: Role Commands
       | document | { "actions": { "update": true } } |
     When I run the command "role:export" with:
       | flag | --path | ./dump |
-    And I delete the role "teacher"
-    And I delete the role "student"
-    When I run the command "role:import" with:
```
Aschen, line 15: Same comment about `import` command

features/User.feature
```diff
@@ -14,16 +14,6 @@ Feature: User Commands
       | flag | --path    | ./dump     |
       | flag | --exclude | gordon     |
       | flag | --exclude | test-admin |
-    And I delete the user "kleiner"
-    And I delete the user "alyx"
-    When I run the command "user:import" with:
```
Aschen, line 19: Same comment about `import` command

src/commands/export/user.ts
```diff
@@ -146,6 +148,30 @@ Examples:
...
+    if (this.flags["mappings"]) {
+      await this._dumpUserMappings();
+    }
+
     return users;
   }
+
+  async _dumpUserMappings() {
+    const filename = path.join(
+      this.flags.path,
```
Aschen, line 160: You should create the directory in case it does not exists

src/hooks/command_not_found/api-action.ts
```diff
@@ -1,31 +1,8 @@
 import chalk from "chalk";
 import { Hook } from "@oclif/config";
-import SdkQuery from "../../commands/sdk/query";
 
 /**
- * Hooks that run the corresponding API action with sdk:query.
```
Aschen, line 6: Actually I would like to keep the inference part which is quite handy when executing API action.  
The automatic inference was more about `kourou document:get` => `kourou api document:get` but then it's more practical to enter common API argument without the `-i` or `-c` (etc) flags

test/commands/app/start-services.test.ts
```diff
@@ -4,11 +4,11 @@ import { execSync } from "child_process";
 const SECOND = 1000;
 const wait = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms));
 
-xdescribe("app:start-services", () => {
+xdescribe("app:services:start", () => {
   test
     .timeout(50 * SECOND)
     .stdout({ print: false })
-    .command(["app:start-services"])
+    .command(["app:services:start"])
```
Aschen, line 11: Shouldn't it be only `services:start` ?

Aschen (changes requested):

src/hooks/command_not_found/api-action.ts
Aschen: @OlivierCavadenti

OlivierCavadenti: I revert this file but I don't understand what to do with this issue at the end : "the automatic inference of commands as API action should be removed"

Aschen: I don't remember why we wrote this when we did the workshop for Kourou v1 but I'm using this inference every day and it's quite handy so if we really want to remove it then I want to discuss first

rolljee (approved):

Aschen:
@OlivierCavadenti Maybe we could remove cucumber and use Jest instead for the functional tests. I think it will simplify the testing

rolljee (approved):

PR #166, opened 2022-10-28
  add support `--print-raw` to `sdk:query`

sebtiz13:
## What does this PR do?
Add flag `--print-raw` to command `sdk:query` to get only the result of query.
It's usefull to use the result of command in pipe with other CLI (like [jq](https://stedolan.github.io/jq/) for exemple)

### How should this be manually tested?

Just download this branch and try
`bin/run sdk:query index:list --print-raw`

The expected result should look like
```json
{
  "indexes": [
    "platform"
  ]
}
```

PR #167, opened 2022-11-02
Add eslint-plugin-kuzzle

OlivierCavadenti:
- Add eslint plugin


Aschen:
I'm not sure this will works with Oclif nazi policies

OlivierCavadenti:
> It's because when `eslint` is installed as a dependency of a dependency it does not create the link for the `eslint` binary.
> 
> I will try to fix this

Seems @sebtiz13 fix that in his PR that can be merged before the others: https://github.com/kuzzleio/kourou/pull/166

Aschen:
I'm not sure I can fix it. I think it's better to install `eslint` package as well. I updated the documentation of the plugin

PR #168, opened 2022-11-16
Add `collection:migrate` command

Aschen:
## What does this PR do?

Adds a command to migrate documents from a dump created with Kourou.

The command takes a script to migrate the content of a document and a path to a collection dump.

Example of migration function
```js
// save this in "migration.js"

function migrate(document) {
  const [firstName, lastName] = document.body.name.split(' ');

  return {
    _id: document._id,
    body: {
      firstName,
      lastName,
    }
  };
}
```

Then run `kourou collection:migrate ./migration.js path/to/dump/`


OlivierCavadenti (approved):
LGTM but update README (add to commands list + doc)

rolljee (approved):

PR #171, opened 2023-01-05
Rework es:migrate to allow dump/restore using files

alexandrebouthinon:
# Why?

For now we always used Elasticdump/Multielasticdump to dump or restore an Elasticsearch cluster content but since this tool do not handle properly our internal indices (ej: `%kuzzle.roles`), we needed to find another way.

# How?

This PR aim to rework the `es:migrate` command improving code and adding the capability to use File provider as output or input during the process. The provider choose is automatic based on URL parsing/File existing.

This new `es:migrate` command allow us to make this kind of migration now:

```bash
kourou es:migrate  --src ./dump --dest http://localhost:9200/ --reset
# Or using the a file as output
kourou es:migrate --src http://localhost:9200/ --dest ./dump
```

> NOTE: I made limited changes on the API usage to avoid breakings (except the environment variable name for source and destination), keep me in touch if you find any regression

## Boyscouting

I patched the `kourou(...)` function that was used to run kourou commands from functional tests. I don't know how it was working but now it use a simple `child_process`


src/support/migrate/providers/elasticsearch.ts
```diff
@@ -0,0 +1,96 @@
...
+    specifications.settings.write = {
+      wait_for_active_shards: 'all'
+    };
+    await this.client.indices.create({ index, body: specifications });
+  }
+
+  async readDataFirst(index: string): Promise<any> {
+    const { body } = await this.client.search({
+      index,
+      scroll: this.options.scrollDuration,
```
OlivierCavadenti, line 52: Instead of scroll query, Elastic says to use sliced point in time queries that can be run in parallel:
https://www.elastic.co/guide/en/elasticsearch/reference/current/point-in-time-api.html#search-slicing

src/support/migrate/providers/file.ts
```diff
@@ -0,0 +1,93 @@
...
+  private getIndexDataFile(index: string): string {
+    return `${this.path}/${index}.json`;
+  }
+
+  async listIndices(pattern?: string): Promise<string[]> {
+    const data = await fs.readdir(this.path);
+    return data
+      // Only keep specs files (data files are can be missing if index was empty)
+      .filter((file: string) => file.match(/\.specs\.json$/))
+      .filter((file: string) => file.match(pattern as string))
```
OlivierCavadenti, line 29: can be done in one filter function

src/support/migrate/providers/elasticsearch.ts
alexandrebouthinon: Good catch, I'm creating an issue to test and replace scroll by Point in Time usage

alexandrebouthinon: https://github.com/kuzzleio/kourou/issues/172

rolljee (changes requested):
Missing functionals tests (I know it will be most likely a pain to test, but this is important)
Also, I am going this afternoon to test it locally so I can have a better overview of this feautre :+1: 

Code looks okay to me tho :+1:

alexandrebouthinon:
> Missing functionals tests (I know it will be most likely a pain to test, but this is important) Also, I am going this afternoon to test it locally so I can have a better overview of this feautre 👍
> 
> Code looks okay to me tho 👍

No problem, you will understand that I do not test the migration from one Elasticsearch to another 😅

PR #173, opened 2023-01-09
kourou 0.25.0

rolljee:
## What does this PR do?

* app:doctor
* es:aliases:cat
* paas:logs
* scroll-ttl in index:export
* es:migrate can now write exported data into files


rolljee:
> Could we wait for this PR #171 ?

yea sure it is ! Any ETA on when this could be merged in this branch ?
I would like to deploy a new version of the paas that will break the paas:logs command

alexandrebouthinon (approved):

Aschen (changes requested):
This should be done with the release tool

rolljee:
you are right

PR #178, opened 2023-01-24
Fix some issues when error occurs

rolljee:
## What does this PR do?

This PR fix issues when the logs method does not return anything that can be displayed

* Error handling when args are put in the wrong order
* Error handling when the server return an internal server error
* Error handling when the server return other than json

### DIFF
--- a/src/support/kuzzle.ts
+++ b/src/support/kuzzle.ts
@@ -227,12 +227,10 @@ export class KuzzleSDK {
 
     // Send the request
     return new Promise((resolve, reject) => {
-      const req = this.ssl ?
-        https.request(url, options, (res) => {
-          resolve(res);
-        }) : http.request(url, options, (res) => {
-          resolve(res);
-        });
+      const module = this.ssl ? https : http;
+      const req = module.request(url, options, (res) => {
+        resolve(res);
+      })
 
       req.on("error", reject);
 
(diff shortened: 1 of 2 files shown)


src/support/kuzzle.ts
```diff
@@ -227,9 +227,12 @@ export class KuzzleSDK {
 
     // Send the request
     return new Promise((resolve, reject) => {
-      const req = http.request(url, options, (res) => {
-        resolve(res);
-      });
+      const req = this.ssl ?
```
Aschen, line 230: I see we resolved the same problem :smile: 

Thus, instead of repeating the code, you can just assign the module before using it `const httpModule = this.ssl ? https : http;`

Aschen (changes requested):

src/support/kuzzle.ts
rolljee: Didn't see that error comming while we developed this :sweat_smile: 

Yea sure, I'll update this is not elegant coding I admit, I did it a little to fast

src/support/kuzzle.ts
```diff
@@ -227,9 +227,10 @@ export class KuzzleSDK {
 
     // Send the request
     return new Promise((resolve, reject) => {
-      const req = http.request(url, options, (res) => {
+      const module = this.ssl ? https : http;
```
Aschen, line 230: This is shadowing the Node.js global `module` variable

Aschen (changes requested):

OlivierCavadenti (approved):

Aschen (approved):

PR #179, opened 2023-01-24
Fix `*:restore` commands and `paas:logs`

Aschen:
## What does this PR do?

The `*:restore` commands where not using the strict mode of `document:mCreate` which can lead to unexpected behavior since errored documents were just silently skipped.  
I also fixed the dumping of errors in a separate file when it's the case.

The `paas:logs` command need to use the `https` module when making HTTPS requests to the PaaS Console


rolljee (approved):

Aschen:
I merged on the wrong branch, I manually reverted this and pushed on `develop` instead

PR #188, opened 2023-06-09
Add command for dumping an Elasticsearch instance in the PaaS

Kuruyia:
## What does this PR do?
This adds the `paas:elasticsearch:dump` command that allows PaaS clients to easily dump the indexes and documents from an Elasticsearch instance of a PaaS application. This PR follows the addition of the `application/storage` controller in the PaaS console.

This command asks for the PaaS environment name and application ID (a project name can also be specified), and a directory where to dump the Elasticsearch data.
The directory structure is automatically created.

First, the indexes are dumped using the `application/storage:getIndexes` action, and are stored in the `<dump_dir>/indexes.json` file.

Then, the documents are dumped usind the `application/storage:dumpDocuments` action, and are stored in the `<dump_dir>/documents.jsonl` file. This file is a [Newline delimited JSON](https://ndjson.org/) file.
Because there could be a huge amount of documents totaling a big size for the response, the PaaS console only allows to dump at most a certain amount of documents at a time. Kourou makes multiple requests to the PaaS console until there are no more documents to dump to get them all. The amount of documents dumped can be set with the `batch-size` flag, but keep in mind that the PaaS console may also limit the maximum amount of documents that can be returned at once.

Finally, the dump session is cleanly terminated and the user is informed where to find their dump files.

image

### Other changes

Types for Node.js were updated to Node 18.

### DIFF
--- a/src/commands/paas/elasticsearch/dump.ts
+++ b/src/commands/paas/elasticsearch/dump.ts
@@ -26,7 +26,7 @@ type DocumentDumpHit = {
   sort: string[];
 };
 
-class PaasInit extends PaasKommand {
+class PaasEsDump extends PaasKommand {
   public static description = "Dump data from the Elasticsearch of a PaaS application";
 
   public static flags = {
@@ -182,4 +182,4 @@ class PaasInit extends PaasKommand {
   }
 }
 
-export default PaasInit;
+export default PaasEsDump;
--- a/src/commands/paas/elasticsearch/dump.ts
+++ b/src/commands/paas/elasticsearch/dump.ts
@@ -139,34 +139,39 @@ class PaasEsDump extends PaasKommand {
     let dumpedDocuments = 0;
     let totalDocuments = 0;
 
-    // Dump the first batch
-    let result = await this.dumpDocuments(pitId, searchAfter);
-    let hits = result.hits.hits;
+    try {
+      // Dump the first batch
+      let result = await this.dumpDocuments(pitId, searchAfter);
+      let hits = result.hits.hits;
 
-    while (hits.length > 0) {
-      // Update the PIT ID and the cursor for the next dump
-      pitId = result.pit_id;
-      searchAfter = hits[hits.length - 1].sort;
+      while (hits.length > 0) {
+        // Update the PIT ID and the cursor for the next dump
+        pitId = result.pit_id;
+        searchAfter = hits[hits.length - 1].sort;
 
-      // Save the document
-      await fs.writeFile(path.join(this.args.dumpDirectory, "documents/", `${currentDocumentChunk++}.json`), JSON.stringify(hits));
+        // Save the document
+        await fs.writeFile(path.join(this.args.dumpDirectory, "documents/", `${currentDocumentChunk++}.json`), JSON.stringify(hits));
 
-      dumpedDocuments += hits.length;
-      totalDocuments = result.hits.total.value;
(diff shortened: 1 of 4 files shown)


src/commands/paas/elasticsearch/dump.ts
```diff
@@ -0,0 +1,185 @@
...
+
+type DocumentDumpHitsTotal = {
+  value: number;
+};
+
+type DocumentDumpHit = {
+  sort: string[];
+};
+
+class PaasInit extends PaasKommand {
```
rolljee, line 29: Classname seems odd

src/commands/paas/elasticsearch/dump.ts
```diff
@@ -0,0 +1,185 @@
...
+    await this.initPaasClient({ apiKey });
+
+    const user = await this.paas.auth.getCurrentUser();
+    this.logInfo(
+      `Logged as "${user._id}" for project "${this.flags.project || this.getProject()
+      }"`
+    );
+
+    // Create the dump directory
+    await fs.mkdir(this.args.dumpDirectory, { recursive: true });
```
rolljee, line 70: does option { recursive true } works as mkdirp ?

rolljee (changes requested):
LGTM, but there are some modifications

src/commands/paas/elasticsearch/dump.ts
Kuruyia: Yes, it works the same as `mkdir -p`

src/commands/paas/elasticsearch/dump.ts
Kuruyia: Yup, that's a copy-paste error! 😅

Aschen:
Why using multiple files instead of a single file in JSONL format?
We use this format with a stream for the index/collection dump to avoid high memory usage.

I see some potentials problems with the actual design:
- having multiple file can end up with literally thousands of small files which is not efficient
- 10 documents limit is really low and the overhead of a network request should be considered, for instance if you have 100000 documents then you will need 10 thousands requests. This limit should be configurable

Actually most of the issues you will encounter here were already discussed and solved for the index/collection export commands. You can have a look how it was done

Kuruyia:
> Why using multiple files instead of a single file in JSONL format?

I didn't know about that file format, I will look at that 😄 

> * 10 documents limit is really low and the overhead of a network request should be considered, for instance if you have 100000 documents then you will need 10 thousands requests. This limit should be configurable

You're right, I think I just forgot to make this limit configurable 😅 

> Actually most of the issues you will encounter here were already discussed and solved for the index/collection export commands. You can have a look how it was done

👍

src/commands/paas/elasticsearch/dump.ts
```diff
@@ -0,0 +1,185 @@
...
+      this.logInfo(`Dumping Elasticsearch documents: ${Math.floor(dumpedDocuments / totalDocuments * 100)}% (${dumpedDocuments}/${totalDocuments})`);
+
+      // Dump the next batch
+      result = await this.dumpDocuments(pitId, searchAfter);
+      hits = result.hits.hits;
+    }
+
+    // Finish the dump
+    try {
+      await this.finishDump(pitId);
```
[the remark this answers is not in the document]
Kuruyia: Good catch, yes it should totally be called.

rolljee (approved):

PR #198, opened 2023-11-23
Allow typescript file to be used by sdk:execute

rolljee:
## What does this PR do?

This PR adds the possibility to use typescript file with the method sdk:execute

  - Step 1 : npm run build
  - Step 2 : ./bin/run sdk:execute < some-typescript-file.ts
  - Step 3 :  The code from your ts file is executed
  ...

### DIFF
--- a/src/commands/sdk/execute.ts
+++ b/src/commands/sdk/execute.ts
@@ -131,7 +131,7 @@ ${this.getVariables()}
     let result;
     try {
       // eslint-disable-next-line no-eval
-      result = vm.runInContext(this.code, vm.createContext({}));
+      result = eval(this.code);
     } catch (error: any) {
       userError = error;
     }
--- a/src/commands/sdk/execute.ts
+++ b/src/commands/sdk/execute.ts
@@ -131,7 +131,7 @@ ${this.getVariables()}
     let result;
     try {
       // eslint-disable-next-line no-eval
-      result = eval(this.code);
+      result = await eval(this.code);
     } catch (error: any) {
       userError = error;
     }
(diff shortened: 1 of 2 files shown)


src/commands/sdk/execute.ts
```diff
@@ -69,33 +71,70 @@ Other
...
+        const file = fs.readFileSync(jsFile, "utf8");
+        this.code = file;
+
+        // Clean up
+        fs.unlinkSync(tsFile);
+        fs.unlinkSync(jsFile);
+      }
+    }
+
+    console.log(this.code);
```
fmauNeko, line 106: Remove log

src/commands/sdk/execute.ts
```diff
@@ -69,33 +71,70 @@ Other
 
   static readStdin = true;
 
+  async toJavascript(filename: string): Promise<boolean> {
+    return new Promise((resolve, reject) => {
+      exec(
```
fmauNeko, line 76: It might be better to use the programmatic api in order to avoid relying on `exec` and to save up on temp file reading / writing.
https://github.com/microsoft/TypeScript/wiki/Using-the-Compiler-API#a-simple-transform-function

src/commands/sdk/execute.ts
```diff
@@ -69,33 +71,70 @@ Other
...
+          // We do not care for TS errors, user should now if this will work or not
+          resolve(true);
+        });
+    });
+  }
+
   async beforeConnect() {
     this.code = this.stdin || this.args.code || "// paste your code here";
+    try {
+      eval(this.code);
```
fmauNeko, line 88: If possible, it would be preferable to attempt evaluating in a separate V8 context
https://nodejs.org/api/vm.html#vmruninnewcontextcode-contextobject-options

fmauNeko (changes requested):

fmauNeko (approved):

PR #204, opened 2024-01-24
Add restore command inside kourou

rolljee:
# Overview

* Add the possibility to restore a snapshot created by kourou, this was missing and should be available


alexandrebouthinon (approved):
Maybe we could add a `pattern` flag to allow users to restore only matching indices

Kuruyia (approved):

rolljee:
> Maybe we could add a `pattern` flag to allow users to restore only matching indices

Indeed, we could add that in another version :)

